PR #49, opened 2022-09-30
Update .Path to .File.Path to fix deprecation warnings.

### LINKED ISSUE #48: .Path when the page is backed by a file is deprecated and will be removed in a future release.
Hi, I saw there was another issue referencing this: #22 
It looks like the hugo version was upgraded, but the warning still remains.

I found a fix here that works: https://github.com/wowchemy/wowchemy-hugo-themes/issues/2592

I'll open a PR with this fix in a bit, I just have to edit two files. 
https://github.com/wowchemy/wowchemy-hugo-themes/issues/2592#issuecomment-1054575549

miodzie:
Closes #48

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -27,7 +27,7 @@
         var liked_article = false
 
         if (typeof auth !== 'undefined') {
-          var oid = "views_{{ .Path }}"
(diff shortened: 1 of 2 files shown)


nunocoracao:
Thanks @miodzie

PR #76, opened 2022-10-04
Remove .DS_Store files

Coronon:
While they are excluded in the `.gitignore`, some `.DS_Store` files have managed to sneak in. This PR removes them. I noticed them because they end up inside my build output.

Thank you for your awesome work!


nunocoracao:
Thanks @Coronon missed this, also updated the .gitignore file to make sure subfolders are considered.

PR #96, opened 2022-10-11
Add pl.yaml Polish Translation

kszpieg:
Hey!

Recently I found this theme and I saw that there is a polish translation missing. I wrote this, maybe it will be useful for someone.


nunocoracao:
Thanks @kszpieg, I'll review this

PR #100, opened 2022-10-13
Ko-Fi icon and link capability

gabrielgbs97:
Added Ko-Fi svg icon with automatic color filling capability, as other FontAwesome 6.x


nunocoracao:
Thanks @gabrielgbs97

PR #112, opened 2022-10-15
Improve hero image visuals

chrisbanes:
- Increase resized image. 600px was too low for a full width hero
- Added rounded corners
- Increased the height, by allowing the img to wrap it's contents (up until 50% viewport height)

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -348,9 +348,8 @@ body:has(#menu-controller:checked) {
 }
 
 .single_hero {
-  background-repeat:no-repeat;
-  background-size:cover; 
-  background-position:center;
+  max-height: 50vh;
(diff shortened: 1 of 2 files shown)


nunocoracao:
Hello @chrisbanes, I don't really like this version. However, I get that some users like you would like to customize the hero image in a different way. Proposal: make the current solution into a partial that can be overridden using hugo normal systems. That way I can control a default style which is more aligned with the theme and it will be way easier to customize.

nunocoracao:
Meanwhile, created the following issue
https://github.com/nunocoracao/blowfish/issues/113

nunocoracao:
Solved in https://github.com/nunocoracao/blowfish/pull/115

PR #121, opened 2022-10-16
added multiple options to display hero images

nunocoracao:


nunocoracao:
@chrisbanes I think you will like this one

chrisbanes:
Nice, looks great!

PR #129, opened 2022-10-16
Optionally disable footer menu. Enabled by default

madoke:
The theme is rendering by default a footer menu with the tags and categories page which cannot be disabled. Currently the only way to hide it is to declare an empty list `footer: []` in the `menus.en.yml`, but the `<nav>` and `<ul>` elements still get renderered

This PR adds a config option to skip the menu when rendering the footer


nunocoracao:
thanks @madoke. This seems like an overkill solution, however, there is a problem with the theme. Hugo will read the defaults for the menu variables from the theme if they are not set in the site config files. The default is being displayed because that is what is configured in the theme default menu.en.toml... I've fixed it in the following commit
https://github.com/nunocoracao/blowfish/commit/82a36a462e38c3032453b7f00333601670c4288a
not only for footer but for the rest of the menu also. You can now simply delete the footer if you don't want to use it.

nunocoracao:
Hello @madoke, re-opening this one. Could you add the documentation to the variable you created? and if possible add it with a default value and commented to both config files (root dir, and exampleSite)?

nunocoracao (changes requested):
Have you considered:
1. removing the parameter from the params.toml default config file
2. Adding the default to the IF clause e.g. if .Site.Params.footer.showMenu | default true
3. documenting the new variable in the respective docs section for theme configurations

madoke:
> 1. removing the parameter from the params.toml default config file

Does this make sense ? What about the others that are already there (`showCopyright`, `showThemeAttribution`, etc..)

> 2. Adding the default to the IF clause e.g. if .Site.Params.footer.showMenu | default true

https://github.com/nunocoracao/blowfish/pull/129/commits/4a8c23c9d591df47cf6856249ccd150322373ca0

> 3. documenting the new variable in the respective docs section for theme configurations

https://github.com/nunocoracao/blowfish/pull/129/commits/2211ce50be3fb42ba9168ff6c99f621dbe608782

config/_default/params.toml
```diff
@@ -16,6 +16,7 @@ enableCodeCopy = true
 # robots = ""
 
 [footer]
+  showMenu = true
```
madoke, line 19: Should I delete this one ? what about the others ?

madoke: Maybe we can comment it out instead

```suggestion
 # showMenu = true
```

nunocoracao (approved):

nunocoracao (approved):

PR #167, opened 2022-10-29
fix: broken link and add: user website

insidemordecai:
Small change, the link under the Users page redirects to what I believe was a previous branch that is now non-existent

Also, added my personal site to the list.


nunocoracao:
Thanks @insidemordecai

PR #203, opened 2022-11-06
improved Japanese translation

kame404:
there are a few strings I saw that I thought could be expressed a little more naturally.


nunocoracao:
Thank you so much @kame404, really appreciate the help :)

PR #246, opened 2022-11-14
Allow overwriting of sharing links

chrisbanes:
This PR changes this by grouping the links within an object key: `Site.Data.sharing.links`. This means that sites can provide their own `data/sharing.json` file and overwrite the `links` key, to provide links of their own links.

Without the key, any custom links in `data/sharing.json` are appended, due to Hugo de-duping keys (not files).

### DIFF
--- a/layouts/partials/sharing-links.html
+++ b/layouts/partials/sharing-links.html
@@ -1,8 +1,7 @@
 {{ with .Params.sharingLinks | default (.Site.Params.article.sharingLinks | default false) }}
-  {{ $links := site.Data.sharing }}
   <section class="flex flex-row flex-wrap justify-center pt-4 text-xl">
-    {{ range . }}
-      {{ with index $links . }}
+    {{ range $.Site.Data.sharing.links }}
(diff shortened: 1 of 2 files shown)


layouts/partials/sharing-links.html
```diff
@@ -1,8 +1,7 @@
 {{ with .Params.sharingLinks | default (.Site.Params.article.sharingLinks | default false) }}
-  {{ $links := site.Data.sharing }}
   <section class="flex flex-row flex-wrap justify-center pt-4 text-xl">
-    {{ range . }}
-      {{ with index $links . }}
+    {{ range $.Site.Data.sharing.links }}
```
chrisbanes, line 3: I don't think if there's a huge difference between this and `site.Data.sharing`, but only `$.Site.Data.sharing.links` worked for me. This also seems to be the [recommended way](https://gohugo.io/templates/data-templates/) to access data.

nunocoracao:
Thanks @chrisbanes

PR #328, opened 2022-12-16
♻️ Add i18n for article series

sergiod26:
Replacing static text for `articles series` with entries in i18n. 

I'm using English for all i18n resources except Spanish (the only other language I know), because I don't like automatic translations as I can't verify if it's a good one. I know is not ideal, but other languages will keep showing as they were before this change


nunocoracao:
thanks @sergiod26

PR #329, opened 2022-12-17
💄 Use grid rather than flexbox for recent articles

chrisbanes:
Currently the items include some inner padding, which means that the items have some leading spacing on the left/right which don't align with the keyline.

This PR updates this to use a grid, with a fixed number of columns depending on the screen size.

Before: (notice that the item doesn't align with the 'recent' word)

Screenshot 2022-12-17 at 16 00 42

With PR:

Screenshot 2022-12-17 at 16 01 17

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -1446,6 +1446,10 @@ select {
   display: flex;
 }
 
+.grid {
+  display: grid;
+}
+
 .hidden {
   display: none;
 }
@@ -1526,10 +1530,6 @@ select {
   width: 100%;
 }
 
-.w-36 {
-  width: 9rem;
-}
-
 .w-24 {
   width: 6rem;
 }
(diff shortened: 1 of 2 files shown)


nunocoracao:
@chrisbanes thanks for this one also. May I ask you if this is something you would like to extend to the rest of list sections across the code. Or if this is only happen for recents (sorry didn't have time to check).

chrisbanes:
Hey, sorry for the late reply! I'll take a look at the other layouts over the holidays and see if anything of them would benefit from moving to a CSS Grid instead.

nunocoracao:
Thanks @chrisbanes meanwhile I merged the PR. Do you mind opening a new one? Thanks

chrisbanes:
Yep, will do!

nunocoracao:
Hey @chrisbanes just noticed that on mobile this change completely eliminates the margins. Think I will need to revisit this.

chrisbanes:
Looks good to me. The extraneous margins were actually what I was trying to remove.

image

nunocoracao:
On the full width option the list gets glued to the screen.

PR #360, opened 2022-12-31
:sparkles: Support for UTF-8 encoded series names; small updates

pfrogond:
As I write in czech, I noticed that when I use UTF-8 encoded characters (š,ř,ž etc.) in a series name, it won't associate it with any articles. The problem is in the _urlize_ function in the _series.html_ and _series-closed.html_ files.

`{{ range $post := sort (index .Site.Taxonomies.series (index .Params.series 0 | urlize)) "Params.series_order" }}`

Besides converting the name lowercase and replacing spaces with hyphens it also changes non ASCII characters to URL friendly format (Čeština -> %C4%8Ce%C5%A1tina). That leads to the inability of the next function: `{{ if eq $post.Permalink $.Page.Permalink }}` to match series names with corresponding posts.

So I changed the code so that now it only converts the name to lowercase and replaces spaces with hyphens.

I also added my personal website, changed one word in the czech translation and added an option to depluralize list names in _config.toml_ file. The last change is useful for non english languages because the pluralize funcition messes up names in other languages.

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -6,6 +6,8 @@
 # baseURL = "https://your_domain.com/"
 defaultContentLanguage = "en"
 
+# pluralizeListTitles = "true"
+
 enableRobotsTXT = true
 paginate = 20
 summaryLength = 30
--- a/exampleSite/content/users/index.md
+++ b/exampleSite/content/users/index.md
@@ -53,6 +53,7 @@ Real websites that are built with Blowfish.
 | [sdehm.dev](https://sdehm.dev)                                        | Personal site                |
 | [dizzytech.de](https://dizzytech.de)                                  | Personal site                |
 | [blog.rotrixx.eu](https://blog.rotrixx.eu)                            | Personal site                |
+| [hexwiki.cz](https://hexwiki.cz/)                                     | Personal site                |
 
 
 
--- a/i18n/cs.yaml
+++ b/i18n/cs.yaml
@@ -18,7 +18,7 @@ article:
   likes:
     one: "{{ .Count }} líbí se mi"
     other: "{{ .Count }} líbí se mi"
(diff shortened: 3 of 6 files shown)


nunocoracao:
Thanks @pfrogond looks good. Would never caught the `urlize` issue as my blog is `eng` only for now. Just need to understand the `pluralizeListTitles` param you added to the `config.toml` and we're good to go.

pfrogond:
The `pluralizeListTitles` is a Hugo function https://gohugo.io/getting-started/configuration/#pluralizelisttitles

Fine for english but poor for pretty much anything else (at least by default,  maybe there is a way to translate it but I find it easier to just disable it)

pfrogond:
Sorry to add another one, but I just came across this thing. The 200 word limit to display the _scroll to top arrow_ `{{ if and (.Site.Params.footer.showScrollToTop | default true) (gt .WordCount 200) }}` in `baseof.html` caueses problems when there are articles with many images but few words. You have a long page but don't get the arrow because the limit hasn't been reached.

pfrogond:
Sorry to add another one, but I just came across this thing. The 200 word limit to display the scroll to top arrow {{ if and (.Site.Params.footer.showScrollToTop | default true) (gt .WordCount 200) }} in baseof.html caueses problems when there are articles with many images but few words. You have a long page but don't get the arrow because the limit hasn't been reached.

I left the limit on 1 so that the arrow isn't displayed on home page, list pages etc. where there is no word counting.

rpstreef:
I'm getting additional padding on the homepage and other list views when I leave the scrollToTop enabled with the default wordcount. I had to set it to 300 to avoid this padding from happening.

nunocoracao:
@rpstreef thanks for the comment, could you please open a new ticket or PR?
I'll look at it for the next release

PR #371, opened 2023-01-02
:adhesive_bandage: :wheelchair: Fix missing title attribute from search button

MilesPernicious:
This was caused by an extra space before an enquoted identifier, causing the i18n lookup to fail.


nunocoracao:
Thanks @MilesPernicious

PR #377, opened 2023-01-03
:sparkles: Add sub-navigation and nested menus 

nunocoracao:


nunocoracao:
@madoke opened this branch/PR for us to work on subnavigation

nunocoracao:
@madoke check this https://tailwindui.com/components/marketing/elements/headers

nunocoracao:
@madoke synced and fixed conflicts... need to check the css. 
question: did you implemented the sub-navigation for the mobile menu? 
note: I've added a parent menu to this brach to make it easier to test before we merge into dev

madoke:
> question: did you implemented the sub-navigation for the mobile menu? 
Nope, that still needs to be worked, the approach will likely be different, maybe we can do an accordion style menu for mobile

nunocoracao:
@madoke wrapped it up. 
- removed javascript all done with CSS only
- added support for mobile
- reviewed the UI/UX of it all
- fixed a bug with icons in the main menu 

Can you please test this using the preview link in this thread and have a read at the docs before I merge in `dev`

docs: https://blowfish-21fff--preview-6w52dqfi.web.app/docs/getting-started/#nested-menus

layouts/partials/header/header-mobile-option-nested.html
```diff
@@ -0,0 +1,30 @@
...
+        {{ if .Pre }}
+        <span {{ if and .Pre .Name}} style="margin-right:3px;" {{ end }}>
+            {{ partial "icon.html" .Pre }}
+        </span>
+        {{ end }}
+        <p class="text-sm font-small text-gray-500 hover:text-gray-900" title="{{ .Title }}">
+            {{ .Name | markdownify | emojify }}
+        </p>
+    </a>
+</li>
```
madoke, line 29: This block could be replaced by either header item simple or header item to reuse code. Header item simple would allow 1 level of nesting, header item would theoretically allow infinite nesting

layouts/partials/header/header-option-nested.html
```diff
@@ -0,0 +1,69 @@
...
+        </a>
+        {{ end }}
+      </div>
+    </div>
+  </div>
+</div>
+
+
+
+<!--div class="relative">
```
madoke, line 38: Maybe we can delete this

layouts/partials/header/header-option-nested.html
```diff
@@ -0,0 +1,69 @@
...
+    </a>
+    <span>
+      {{ partial "icon.html" "chevron-down" }}
+    </span>
+  </div>
+  <div class="absolute menuhide">
+    <div class="pt-2 p-5 mt-2 rounded-xl backdrop-blur shadow-2xl">
+      <div class="flex flex-col space-y-3">
+        {{ range .Children }}
+        <a href="{{ .URL }}" {{ if or (strings.HasPrefix .URL "http:" ) (strings.HasPrefix .URL "https:" ) }}
```
madoke, line 19: Same as the comment above ? Could we try to reuse the header menu simple ?

layouts/partials/header/header-mobile-option-nested.html
nunocoracao: Nope, css classes are different

layouts/partials/header/header-option-nested.html
nunocoracao: Maybe we can 🤣

layouts/partials/header/header-option-nested.html
nunocoracao: Doesn't have the href

layouts/partials/header/header-option-nested.html
nunocoracao: done

nunocoracao:
I don't think there is any value in reusability plus the entries you mentioned are different (CSS, href in a for nested option, etc) I tried to have that into account when creating them. Will leave this one for last but going through the current bug list. Nest step is to remove examples and merge to dev.

nunocoracao:
@madoke recovered sub-navigation.... sorry please check the preview link and test everything.

madoke (approved):

PR #407, opened 2023-01-08
:bug: added support for heroStyle via front-matter

adir1:
This is to support scenarios when custom heroStyle is specified in a single article's front-matter

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -3,7 +3,11 @@
 
 <article>
   {{ if .Params.showHero | default (.Site.Params.article.showHero | default false) }}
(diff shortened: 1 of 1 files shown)


nunocoracao:
Thanks, @adir1 I'll accept this one but I will need to refactor the code if you don't mind. I can make it a little cleaner if I have time.

nunocoracao:
not just refactor... but want to keep a overide logic. EVEN if the user is defining a heroStyle for all articles, it should be possible to override for specific ones.

adir1:
I appreciate it and totally agree. I tried to do it with inline-If at first, but apparently that feature doesn't exist in Go lang. But hopefully there is something nicer (in Go or Hugo) - will be happy to learn.

nunocoracao:
Done you can have a look

PR #450, opened 2023-01-16
:sparkles: shortcode gallery

### LINKED ISSUE #428: 🌟 Shortcode: Gallery

In my website, I need the ability to easily showcase multiple related images and I would like a better solution than to just use a bunch of `{{< figure ... >}}` shortcodes.

A `{{< gallery ... >}}` shortcode that would match multiple images given a regex would be a nice solution for this.

Example/Expected usage:

```
# To add all png images in a directory:

{{< gallery source="img/*.png" >}}

# To add all pngs that contain `gallery_` in the name:

{{< gallery source="img/gallery_*.png" >}}
```

N/A

**Additional context**

Useful links:

- https://github.com/Darthagnon/hugo-easy-gallery: this repo implements something very similar, but requires us to put all images in `static/` instead of using page bundles

mirceanton:
Fixes #428

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -167,6 +167,26 @@ Blowfish also supports automatic conversion of images included using standard Ma
 
 {{< figure src="abstract.jpg" alt="Abstract purple artwork" caption="Photo by [Jr Korpa](https://unsplash.com/@jrkorpa) on [Unsplash](https://unsplash.com/)" >}}
 
+## Gallery
+
+`gallery` allows you to showcase multiple images at once, in a responsive manner with more varied and interesting layouts.
+
+In order to add images to the gallery, proceed as you normally would with `{{< figure >}}` shortcodes, but made sure to add the extra `class="grid-item grid-wXY"` in order for the gallery to properly pick the images up. The widths available by default start at 10% and go all the way to 100% in 5% increments. For example, to set the width to 65%, set the class to `grid-w65`.
+
+**Example:**
+
+```md
+{{</* gallery */>}}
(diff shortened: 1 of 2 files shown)


mirceanton:
@nunocoracao For this one, I decided to go with the verbose list as a solution since it makes it that much more easier to set the custom width for each element.

nunocoracao:
@mirceanton I don't agree with this implementation. 

1) can we use tailwind components please, defining styles in shortcodes is going to lead to problems. 
2) why do we need the new import?

mirceanton:
@nunocoracao Okay, I'll see what I can do. Sorry if this isn't following best practices, I don't normally do HTML, CSS & JS, it's all a bit new to me.

mirceanton:
@nunocoracao As for the javascript import, should I simply dump the minified js into a file under `static/js` like we have `zoom.min.js`?

nunocoracao:
For the CSS it will be super clear if you look at other shortcodes or partials. Tailwindcss docs are also super great. 
Pro-tip for dev: run `npm run example` in one terminal window and `npm run dev` in another. The first one runs Hugo server and the second one compiles tailwind in real time if there are any changes. When you finish use `npm run build` to generate the final CSS file for the theme. 

For the imports check what is happening for typeit for example. 
- add a new npm module to devDependencies
- change the package.json copy-to-from array to add the new dependency
- run `npm run fullinstall`
- final lib should be in assets
- change vendors partials to ONLY import the lib if the shortcode is being used (there are other examples there)

Sounds like a lot of effort but it will make everything super easy to update and maintain. My advice, check if you really need a new dependency for this or if could be achieved using just tailwind. Aka a quick search for twailwind image gallery might surprise you.

nunocoracao:
@mirceanton if you want I can open a feature branch an we can collab before merging to dev (release branch)

mirceanton:
@nunocoracao that would be amazing! This is the first time I'm touching this tech stack so I am quite lost to be honest.

nunocoracao:
merging into feature branch for final alterations

PR #458, opened 2023-01-16
✨ shortcode gallery

nunocoracao:


nunocoracao:
@mirceanton new PR with a separate branch for collab

nunocoracao:
@mirceanton did a revamp to the code to polish it. Have a look and let me know. This should be enough for you to do the slider. If I have time during the week I might take a look at it. 

Note: let's wrap this one first before moving to the slider. There will be conflicts once we merge to `dev` so it's better to do one at a time.

exampleSite/content/docs/shortcodes/index.md
```diff
@@ -172,20 +172,30 @@ Blowfish also supports automatic conversion of images included using standard Ma
...
 {{< gallery >}}
-  {{< figure src="abstract.jpg" alt="Abstract purple artwork" caption="Photo by [Jr Korpa](https://unsplash.com/@jrkorpa) on [Unsplash](https://unsplash.com/)" class="grid-item grid-w75">}}
-  {{< figure src="featured.png" class="grid-item grid-w15" >}}
+  <img src="gallery/01.jpg" class="grid-w33" />
+  <img src="gallery/02.jpg" class="grid-w33" />
+  <img src="gallery/03.jpg" class="grid-w33" />
+  <img src="gallery/04.jpg" class="grid-w33" />
+  <img src="gallery/05.jpg" class="grid-w33" />
+  <img src="gallery/06.jpg" class="grid-w33" />
+  <img src="gallery/07.jpg" class="grid-w33" />
```
mirceanton, line 198: This part I am not so sure about. I don't really like the idea of recommending/having to use inline HTML inside markdown. I would much more prefer using `{{< figure >}}` tags instead of raw ``.

I believe it goes against markdown guidelines/lint-rules to use HTML inline and thus I generally try to avoid it.

assets/css/main.css
```diff
@@ -446,4 +446,26 @@ body:has(#menu-controller:checked) {
...
+.grid-w30 { width: calc(30% - 5px); margin: 0px !important; }
+.grid-w33 { width: calc(33% - 5px); margin: 0px !important; }
+.grid-w35 { width: calc(35% - 5px); margin: 0px !important; }
+.grid-w40 { width: calc(40% - 5px); margin: 0px !important; }
+.grid-w45 { width: calc(45% - 5px); margin: 0px !important; }
+.grid-w50 { width: calc(50% - 5px); margin: 0px !important; }
+.grid-w55 { width: calc(55% - 5px); margin: 0px !important; }
+.grid-w60 { width: calc(60% - 5px); margin: 0px !important; }
+.grid-w66 { width: calc(66% - 5px); margin: 0px !important; }
+.grid-w65 { width: calc(65% - 5px); margin: 0px !important; }
```
mirceanton, line 464: Just a pet peeve of mine and maybe I have a bit of OCD, but these 2 are not in the right order 😆

mirceanton (changes requested):

assets/css/main.css
nunocoracao: For crying ou loud... 🤣

exampleSite/content/docs/shortcodes/index.md
mirceanton: I see, okay then. No other comments on this. Thread resolved.

assets/css/main.css
mirceanton: Ah, I tried to fix it myself but apparently I do not have permission to push to this branch.

layouts/shortcodes/gallery.html
```diff
@@ -0,0 +1,18 @@
...
+  {{ .Inner }}
+</div>
+
+<script>
+
+  document.addEventListener("DOMContentLoaded", function () {
+    var elem = document.querySelector('#{{ $id }}');
+    var pckry = new Packery(elem, {
+      percentPosition: true,
+      gutter: 5,
```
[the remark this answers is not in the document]
nunocoracao: Thought about that. Check the CSS though... the gutter needs to be baked in the CSS otherwise what you had defined would break the gallery. E.g. width 25% plus gutter doesn't allow for 4 columns...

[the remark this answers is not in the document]
mirceanton: Ah... Yeah, i missed that. You're right. I've nothing else to add then. Looks good to me

assets/css/main.css
nunocoracao: you can't commit directly to it, but you can fork and create a PR to the branch... but I'll fix it no issues

PR #459, opened 2023-01-16
✨ shortcode carousel

nunocoracao:


nunocoracao:
@mirceanton new PR with a separate branch for collab

nunocoracao:
@mirceanton we'll also work on this one... needs rework

mirceanton:
@nunocoracao definitely! I'll try to get to it tomorrow. I'll try to do this one though. Your last PR gave me enough of an example, I believe, to fix this one myself. I'll let you know if I run into any issues but hopefully it'll be fine

nunocoracao:
@mirceanton sorry already on it... :D

nunocoracao:
wIll just remove the dependency for this one. We can replicate the same thing using tailwind no need to add a dependency for this. I will push my code today and you can pick it up / review.

mirceanton:
You're too fast for me ☺️

nunocoracao:
The preview links for each PR should also be fixed from another PR I merged today. @mirceanton you'll be able to see the results in the preview link

layouts/shortcodes/sliderOLD.html
```diff
@@ -0,0 +1,46 @@
...
+      el: '.swiper-pagination{{ $id }}',
+      clickable: true,
+    },
+  
+    navigation: {
+      nextEl: '.swiper-button-next{{ $id }}',
+      prevEl: '.swiper-button-prev{{ $id }}',
+    },
+  });
+</script>
```
mirceanton, line 46: I believe this file should be removed now. Right?

exampleSite/content/docs/shortcodes/index.md
```diff
@@ -90,6 +90,26 @@ Call to action
...
 {{< /button >}}
 
+## Carousel
+
+`carousel` is used to showcase multiple images in an interactive and visually appealing way. This allows a user to slide through multiple images while only taking up the vertical space of a single one. All images are displayed using the full width of the parent component and using a ratio of `16:9`.
+
+<!-- prettier-ignore-start -->
+| Parameter | Description                                               |
+| --------- | --------------------------------------------------------- |
+| `images`  | This shortcode receives how many image paths as you want. |
```
mirceanton, line 100: I am not so sure about this table as it kinda suggests that there is a named parameter called `images`. I think the examples should suffice.

layouts/shortcodes/carousel.html
```diff
@@ -0,0 +1,38 @@
+<div id="carouselExampleCaptions" class="carousel slide relative" data-bs-ride="carousel">
```
mirceanton, line 1: Here, the ID should either be dynamically generated or user provided. If we leave it like this, we cannot have multiple carousels in the same page, or the controls from either one will just act on the first.

mirceanton:
@nunocoracao See #467 as it fixes all of these comments

layouts/shortcodes/sliderOLD.html
nunocoracao: yes thanks

layouts/shortcodes/carousel.html
nunocoracao: nice one

exampleSite/content/docs/shortcodes/index.md
nunocoracao: Saw your changes thanks

nunocoracao:
merging, thanks again @mirceanton

PR #484, opened 2023-01-25
✨ Custom colors for alert shortcode

### LINKED ISSUE #483: ✨ Customizable alert card

Currently we can only customize the icon for the alert shortcode, not the colors.

I would like the colors for both the card and the icon itself to be customizable. For example, having a red card with a black warning icon for very important things etc.

mirceanton:
Fixes #483

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -26,7 +26,7 @@ In addition to all the [default Hugo shortcodes](https://gohugo.io/content-manag
 
 The input is written in Markdown so you can format it however you please.
 
-**Example 1:**
+**Example 1:** No params
 
 ```md
 {{</* alert */>}}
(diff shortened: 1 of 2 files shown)


mirceanton:
Not sure if the way I tried to implement "optional" parameters is the actual proper way. Looking for some fedback.

nunocoracao:
Can we add a table with all the params to the docs for this?

mirceanton:
Sure! I'll add it tomorrow

mirceanton:
I lied. I did it today!

image

nunocoracao:
I think we can collapse those notes into the table itself thoughts?

mirceanton:
image

Something like this perhaps?

nunocoracao:
@mirceanton u rock

nunocoracao:
Just noticed something, this is a breaking change isn't it?
People using the current Alert shortcode will have problems with this change aka it will stop working.

nunocoracao:
Idea: instead of replacing the old shortcode, why don't we create a new one? I know it sucks in terms of code... but whoever is using an older version won't have their site broken when they update. thoughts?

nunocoracao:
Alternatively, if there is a way to keep the same syntax for icons as before that would be great. I don't think there is but you seem to be more comfortable with the argument passing in partials than I am. 

@mirceanton check above messages when you have time

mirceanton:
@nunocoracao I am almost sure there is a way to make this work. I remember reading something in the hugo docs at some point.

https://gohugo.io/templates/shortcode-templates/#isnamedparams

I think this should do. I'll give it a try

mirceanton:
@nunocoracao I think this should do! It now works both ways.

Not really sure how to cleanup/organize the code. I think those if statements look a bit ugly, but I didn't find a tidier alternative.

PR #490, opened 2023-01-26
📚 Fix default config to match the description from the docs site

### LINKED ISSUE #486: :bug: `article.showEdit ` is `true` in config file when defaults should be `false`
**Describe the bug**

I noticed that the edit button is present on articles on my site even though I didn't enable this. [In the documentation for configuration](https://blowfish.page/docs/configuration/#article) it indicates that `article.showEdit` defaults to `false` however [in this code](https://github.com/nunocoracao/blowfish/blob/main/config/_default/params.toml#L59) I am seeing it set to `true`.

**To Reproduce**

Create a new project and create a post. View the post and the edit button should be present without any changes to configuration.

**Expected behavior**
I would expect the edit button not to be shown since it should be `false` by default.

mirceanton:
Fixes #486


nunocoracao:
@mirceanton thanks for this, you aligned everything to the defaults in the docs right?

mirceanton:
Yep, i did. Feel free to double check me on it, though. I did comment out some "dummy" sections that were "Not set" in the docs

nunocoracao:
I probably will accept this and actually to a review of how I want a "vanilla" installation to look, and then change the docs. This is super helpful thanks

mirceanton:
Glad I could help!

PR #499, opened 2023-01-31
✨ Shortcode `carousel`: add optional parameter for scroll duration

mirceanton:
Currently, if I have multiple `carousel` elements in the same page, they scroll more or less at the same time, which looks kind of bad.

This PR solves the problem by adding a pseudo-random time to each carousel to stagger them a bit, making the page look more "dynamic".

The base duration for scrolling is set to 2000 millis. On top of that, a pseudo-random number from `0` to `999` millis is added, to have random values in the `[2000, 2999]` range.

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -106,7 +106,7 @@ Call to action
 | ------------- | ----------------------------------------------------------------------------------------------------------------- |
 | `images`      | **Required.** A regex string to match image names.                                                                |
 | `aspectRatio` | **Optional.** The aspect ratio for the carousel. Either `16-9`, `21-9` or `32-9`. It is set to `16-9` by default. |
-| `duration`    | **Optional.** The duration for the auto-scrooling, specified in milliseconds. Defaults to `2000` (2s)             |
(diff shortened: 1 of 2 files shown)


nunocoracao:
Sorry @mirceanton I didn't see a ticket for this one. I don't like this solution. Let's move on with making the duration an optional parameter for the shortcut. This would let the user define what behavior they want for their use-case vs randomizing without any good reason for it.

mirceanton:
@nunocoracao Yep, I created this PR without opening an issue first, my bad!

Yeah, in hindsight I think that setting it manually is a better choice than randomizing it. We have a predictable outcome this way. Fair point. I'll get to it

nunocoracao:
Thanks @mirceanton

mirceanton:
@nunocoracao Any opinion/preference on the name of the param?

- `duration`
- `autoScrollDuration`
- `durationMillis`
- `autoScrollDurationMillis`
- `interval`
- etc.

For now I left it as `duration` with a mention in the docs that it is expressed in millis because I found the other options to be unnecessarily long.

nunocoracao:
What does the underlying framework use? Maybe just use that?

mirceanton:
The underlying framework calls it `data-bs-interval`. I'll change it to `interval` then.

Issue #519, opened 2023-02-12
🔬 iOS "Notch" on Safari Ignores Theme Colors


Capprin:
**Describe the bug**
On iOS, in Safari, the notch is a plain white, ignoring dark mode theme colors (or an image background). In landscape mode, _colors_ are respected, but background images do not extend to the edges of the screen.

**To Reproduce**
1. Go to https://blowfish.page on iPhone Safari
2. If in light mode, click moon icon
4. See white bar above theme (and light search bar)
5. In landscape mode, see correct color change but bad background image behavior

**Expected behavior**
Theme content smoothly interacts up to edges of available display; iOS search bar colors respect darkness

**Screenshots**
![bf_portrait](https://user-images.githubusercontent.com/1781067/218326245-7e1628b9-ae6a-4d05-9d96-300a3cc377c9.jpg)
![bf_landscape](https://user-images.githubusercontent.com/1781067/218326253-2c683c54-5998-4a33-916e-949fefbd92ee.png)

**Desktop**
 - OS: iOS 16.1.2
 - Browser: Safari 16

**Hugo & Blowfish versions**
Hugo v0.110.0, but this issue persists even on the public Blowfish site.
Checked out Blowfish main at [55f6cc](https://github.com/nunocoracao/blowfish/commit/55f6cc8873bed9447801168c7ae11df89c34fd00)

**Additional context**
This is a tricky issue to solve. The correct fix will probably be to expand background elements to the edges of iPhones with the relevant [meta tags](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html), and defining a safe region for content in the stylesheet. This is fine with images, but it probably won't fix the underlying coloring problem for other pages because apple appears to choose colors with the [theme-color meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color) before the stylesheets are loaded. That precludes the use of the theme colors defined in the blowfish stylesheets.

My own hotfix just manually updates the theme color metadata at build time using colors defined in the site parameters (ugly, I know). It's technically responsive, but I've had to turn off the "theme switch" button because it doesn't update the meta tag - that might be another thing that needs to happen.

This iPhone-specific issue is annoying to see and annoying to fix. I'm using my "bandaid" for now so I don't need to update the stylesheets or any of the javascript. If this is fixed correctly, it would be fantastic.

nunocoracao:
[...]

PR #520, opened 2023-02-12
Ugly hotfix for iOS notch color

Capprin:
Technically addresses #519 but fails to consider background images and theme switch button.

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -8,6 +8,9 @@
 colorScheme = "blowfish"
 defaultAppearance = "light" # valid options: light or dark
 autoSwitchAppearance = true
+# expose theme colors for iOS notch coloring (barring better options)
+iosLightOverride = "ffffff"
+iosDarkOverride = "1e313b"
 
 enableSearch = false
 enableCodeCopy = false
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -4,6 +4,8 @@
   <meta http-equiv="content-language" content="{{ . }}" />
   {{ end }}
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta name="theme-color" content="#{{ .Site.Params.iosLightOverride | default "ffffff" }}" media="(prefers-color-scheme: light)" />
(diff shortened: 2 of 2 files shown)


nunocoracao:
@Capprin thanks for submitting this.

config/_default/params.toml
```diff
@@ -8,6 +8,9 @@
 colorScheme = "blowfish"
 defaultAppearance = "light" # valid options: light or dark
 autoSwitchAppearance = true
+# expose theme colors for iOS notch coloring (barring better options)
+iosLightOverride = "ffffff"
```
nunocoracao, line 12: delete these, not needed - should use the selected theme color set for the background.

layouts/partials/head.html
```diff
@@ -4,6 +4,8 @@
   <meta http-equiv="content-language" content="{{ . }}" />
   {{ end }}
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta name="theme-color" content="#{{ .Site.Params.iosLightOverride | default "ffffff" }}" media="(prefers-color-scheme: light)" />
```
nunocoracao, line 7: Make use of the already defined classes/colors for tailwind. Much rather have a different HTML implementation to leverage those classes that this+more configs.

Capprin: @nunocoracao maybe I'm missing something obvious (I haven't used CSS in a while, or modern CSS ever), but when I tried to use CSS colors in these meta tags, they weren't populated quickly enough (or just didn't work) for iOS to change the notch color. I agree that this is a terrible workaround, but it works for now barring a more extensive set of changes.

nunocoracao: @Capprin there is people using this with different themes and colors. These feature needs to be seamless otherwise it will lead to a breaking change. This should use one of the colors defined in the theme. Maybe not via a meta tag but doable.

nunocoracao:
@Capprin I will not have time to work on this and I don't think this solution is good. Are you planning to fix my comments? Otherwise I suggest we close this PR and discuss further in the ticket before making code changes.

nunocoracao:
Closing this PR, found a way to do it automatically without requiring configs.
https://github.com/nunocoracao/blowfish/pull/546

PR #593, opened 2023-03-11
♿️ Alt text management

androw:
This change improve accessibility by:
- indicating that the background image is use for presentation only and then do not need an alt text
- add an alt text to the big hero image  defaulting to page title or to the specified "alt" param in the Page Bundle.

I've also set a fixed image ratio for the big hero image to improve performance (less page redraws).


nunocoracao:
Thanks @androw

PR #595, opened 2023-03-12
❇️ Use article/page configured summary ahead of default

madoke:
This PR is a potential small seo improvement. If there is a configured `summary` in the page frontmatter, it will be used in the `<meta name="description">` tag instead. If this isn't configured, then it defaults to `Site.Params.description`


madoke:
I assume that something similar could also be done with the keyworkds

nunocoracao:
Thanks @madoke winner.

PR #622, opened 2023-03-20
Accessibility: add alt tag to carousel

dotlineX:
As mentioned on #617


dotlineX:
Could also add alt tag to background image on the home page

nunocoracao:
Thanks @dotlineX 
ack and ok on that second change you proposed. Accepting this one to keep the PRs nice and small.

Issue #639, opened 2023-03-29
✨ Add Gitlab repo / project shortcode


DiogoM1:
**Is your feature request related to a problem? Please describe.**
I would like to be able to use a shortcode, like the one available for github, to add / reference gitlab repositories.

**Describe the solution you'd like**
Add a gitlab repo based shortcode

**Describe alternatives you've considered**
Using simple links which leads to an inconsistency between Github / Gitlab project presentation.

**Additional context**
I built a small prototype locally and I took the following screenshot.
![image](https://user-images.githubusercontent.com/43378713/228395194-80f3fcce-e005-45c4-b760-6d7e83e7005b.png)

Unfortunately gitlab offers no information regarding programming languages used in a project within the REST API response body.

I would be willing to make a PR based on your feedback to add this functionality. It was pretty simple too.

nunocoracao:
Thanks for https://github.com/nunocoracao/blowfish/pull/641 @DiogoM1 left a comment there regarding docs. Will test it in the dev branch

PR #641, opened 2023-03-29
✨ Add gitlab shortcode

DiogoM1:
#639 PR with prototype implementation of the proposed shortcode.

It is a simple adaptation of the current github shortcode.


nunocoracao:
@DiogoM1 thanks Diogo
Could you please also add this to the docs shortcode page?

DiogoM1:
@nunocoracao Hey, hope it addresses your request. I added a GitLab instance URL as there are many institutes like the one I was at that host their packages in public self hosted instances. Hope the documentation also covers everything fully.

zwebcourses:
hey
template for shortcode "github" not found
how fix that?

PR #649, opened 2023-04-05
fix: Remove heading anchor content

### LINKED ISSUE #648: :bug: [Accessibility] Anchors broken on Brave Speedreader
**Describe the bug**
Headers create an additional anchor element 100px above the header itself to be visible when scrolled into view.
Also it breaks automatically generated descriptions in RSS feeds.

**To Reproduce**
1. Use some headers `#`, `##`, etc.
2. Enable Brave Speedreader
3. See doubled anchors with `>` at the end

**Expected behavior**
Anchor div does not have to have any content, then the issue does not occur.

**Screenshots**
Brave Reader mode
image

RSS feed auto generated description
image

**Desktop:**
 - Browser: Brave

**Hugo & Blowfish versions**
`v2.31.3`

**Additional context**
I will create a PR for this!

maciejopalinski:
closes #648


maciejopalinski:
@nunocoracao any thoughts about this small fix?

PR #660, opened 2023-04-10
🐛 fix card height in lists

nunocoracao:


nunocoracao:
Solves https://github.com/nunocoracao/blowfish/issues/658

PR #683, opened 2023-04-26
✨[#676] Add logo dark/light mode switching mechanism

### LINKED ISSUE #676: ✨ Dark/Light mode-based logo switching
I have two versions of my logo, one which is suitable for dark mode and another which is suitable for light mode.
Having the theme switch between both would look better than trying to create a single logo to fit both appearance modes.

Be able to define a light and a dark mode logo in the theme configuration files. Whenever the user switches between light or dark appearance mode, the image is swapped respectively.

**Additional context**

DanWiseProgramming:
This adds another parameter called `secondaryLogo` which, if set, will be switched to from the default (and now primary) logo. Which logo is primary or secondary is dictated by the `defaultAppearance` parameter.

Any `` tag with class `logo` gets its `src` attribute updated based on appearance mode.
This is useful if the logo appears several times on a single page.

I decided to implement the logo switching mechanism in `appearance.js` as this:

1. does not require user re-generation and override of Tailwind CSS stylesheet.
2. allows the proper use of an `` tag to be retained (as opposed to using background-image and Tailwind CSS to switch the logo images based on appearance mode).
3. retains the current `logo` parameter.
4. retains the default Hugo behaviour of only including images in `public` which are referenced.

If this is accepted as a feature/proposal/solution I am happy to document it.

If you see anything that doesn't look right or could be improved, I would appreciate all and any feedback. Thank you.

Closes #676 .


DanWiseProgramming:
An alternative approach would be to set the paths as `data-logo-path` and `data-secondary-logo-path` attributes on the logo `` tag in `basic.html` and then read that from `appearance.js`. Not sure if that would be a preferred approach to solving this problem?

nunocoracao:
@DanWiseProgramming thanks for the contribution. I'll need time to review this one as it has some implications which I need to test. No issues with the approach at this stage. 

ask: would you mind updating the docs in the same PR? Otherwise no on really knows about the variable and how to use it. Thanks

nunocoracao:
@DanWiseProgramming sorry my bad, this was closed because the target branch was merged. I'll reopen it tomorrow.

nunocoracao (commented):
- add documentation for the new paramemter
- (question) is there a way to opimize the update logo function? I wonder if there is a way to don't have any javascript loaded if there is no secondary logo config'd

DanWiseProgramming:
Hi @nunocoracao ,
I'll add documentation either tomorrow or the day after.
As for optimising the JS function - I'll think about this and get back to you on that. It's a good idea 👍🏻

DanWiseProgramming:
@nunocoracao Hm, just realised that I hadn't committed two of the files which have necessary changes. Sorry about that. Hope the changes are still ok. I'll give this all a test just to make sure I haven't missed any other files.

I'll also work on the tasks above (docs & JS function optimisation).

layouts/partials/header/basic.html
DanWiseProgramming, line 1: This attaches the update logo JS function to the DOM.

DanWiseProgramming:
Hi @nunocoracao,
I think that this PR is ready for a re-review whenever you get the chance. I think that I've addressed both of your review points above.

If I've missed something, please do let me know and I'll try to rectify it. Thank you.

nunocoracao:
thanks again @DanWiseProgramming

DanWiseProgramming:
Thank you for the review @nunocoracao

Issue #681, opened 2023-04-26
✨ Substack icon / link


jpmcb:
Howdy! Love the theme. Incredible. Well done. I'd love if there was a [Substack](https://substack.com) icon for the profile mode in order to point people to my newsletter which i publish there.

I'd be happy to contribute if you'd be able to get me started - thanks much!

nunocoracao:
@jpmcb thanks for the idea, this one is really easy I can do it.

PR #684, opened 2023-04-29
✨ Substack icon

nunocoracao:


nunocoracao:
solves https://github.com/nunocoracao/blowfish/issues/681

Issue #688, opened 2023-04-30
✨ Allow Top-Level Menu Entry with nested sub-menu to actually link to a page


schachem:
**Is your feature request related to a problem? Please describe.**
I have a page bundle called "about" with several leaf entries like education, work, etc.
The list type site for "about" does contain some text.

When creating a nested menu like:
Top-Level: About
Sub-Menu: Work
Sub-Menu: Education

, the Top Level menu entry About does not actually link to the list site, as the partial does not contain a href.

**Describe the solution you'd like**
Rewrite the partial for nested menus such that it contains the href to the Top-Level menu url (it existing).

PR #689, opened 2023-04-30
✨ Fix missing URL for Top-level menu entry in nested menu

schachem:
I hope this is okay.
I am explicitly checking if there even is a url, and if so, include it.

### DIFF
--- a/layouts/partials/header/header-option-nested.html
+++ b/layouts/partials/header/header-option-nested.html
@@ -5,7 +5,8 @@
       {{ partial "icon.html" .Pre }}
     </span>
     {{ end }}
-    <a class="text-base font-medium text-gray-500 hover:text-gray-900" title="{{ .Title }}">
+    <a {{ if .URL }} href="{{ .URL }}" {{ if or (strings.HasPrefix .URL "http:" ) (strings.HasPrefix .URL "https:" ) }}
+      target="_blank" {{ end }} {{ end }} class="text-base font-medium text-gray-500 hover:text-gray-900" title="{{ .Title }}">
       {{ .Name | markdownify | emojify }}
     </a>
     <span>


schachem:
Fixes #688

schachem:
Well, that was some sort of quick shot from my side; just noticed that I violated a few contributing guidelines as I did not really take the time to properly read through them as it was such a minor change.
If you agree to include this change though, I can open another PR and properly adhere to the guidelines :)

nunocoracao:
Thanks for the contribution @schachem. And don't worry about the guidelines, no harm done. A couple of asks though which I hope you don't mind making as part of the same PR

1) update docs 
 i.e. - https://blowfish.page/docs/getting-started/#nested-menus
this part `Note that pageRef and url will be ignored for the parent entry.`

2) Even though I think the behavior will be weird I don't mind supporting this, however, let's do it in one go. i.e. You should also check for `pageRef` and make sure to support both, meaning one OR the other OR none as of today.  

Let me know if it makes sense and if you don't mind updating the PR. 

Thanks again for the help

nunocoracao:
@schachem just checking for an ack, do you agree with the comments?

schachem:
Yes, more or less.

However, I am not sure about the remark regarding `pageRef`.

In my experience, either page ref or a url being set in the menu properties will lead to an .URL param being present or not.

So I think one does not need to check the if pageRef is set.

But I will look into it more thorough

nunocoracao:
@schachem fair point. You might be right don't remember that part of the codebase would need to check. Documentation-wise at least both should be mentioned as supported.

nunocoracao:
@schachem will update the docs myself
Thanks again

nunocoracao:
fixes https://github.com/nunocoracao/blowfish/issues/688

Issue #703, opened 2023-05-04
🐛 Links become white on hover in light theme


dosoft:
When I move my mouse over a link in the text on the website's light theme, the link color changes to white, making it difficult to read against the background. This happens consistently across all pages and text blocks on the website's light theme.

nunocoracao:
Thanks @dosoft, confirmed should be an easy fix. Will pick it up when I have time. Marked as a `good first issue` to see if someone want to open a PR to fix.

MOHDNEHALKHAN:
@dosoft Check my PR I changed the color

nunocoracao:
@dosoft @MOHDNEHALKHAN sorry I closed the PR, it was not changing the right files. Please have a look at the docs on how to update the CSS values and generate the compiled CSS files.

MOHDNEHALKHAN:
@nunocoracao Ok I will check it, as  I am a beginner I don't know much about how to contribute to a large project so if I understand it will definitely create a PR again

PR #707, opened 2023-05-06
Changed hover color for all links in light mode

MOHDNEHALKHAN:
#703 Issue Solved
I have changed the color when hovering over any link it gets white in color which was unable to read so I change the color.


nunocoracao:
Thanks for the contribution @MOHDNEHALKHAN but this will not solve it. I am closing the PR.
You changed the compiled `css` file, not the original configurations for the theme. Example: if I approve this one, my ci/cd pipelines will overwrite that file.

PR #708, opened 2023-05-07
🔧 Fixed zoom.min.js library not being found if site is in a subfolder.

MaikelChan:
zoom.min.js is not found when site is not in the root and baseURL is set accordingly to something like `https://website.com/test/`. Other files in the static folder work fine. Only zoom.min.js has this issue.


nunocoracao:
Thanks for the PR @MaikelChan

PR #731, opened 2023-05-21
Fixed having multiple galleries in the same page, only works the first one

MaikelChan:
Having multiple galleries in the same page makes them have the same ID (as now.UnixNano can be the same), so only the javascript of the first gallery works. This adds a unique number to the ID of each gallery so it fixes the problem.

### DIFF
--- a/layouts/shortcodes/gallery.html
+++ b/layouts/shortcodes/gallery.html
@@ -1,5 +1,5 @@
 {{ $time := now.UnixNano }}
-{{ $id := delimit (slice "gallery" $time) "-" }}
+{{ $id := delimit (slice "gallery" .Ordinal $time) "-" }}
 
 <div id="{{ $id }}">
   {{ .Inner }}


MaikelChan:
Also, I've seen that there have been reports in the past related to gallery images being overlapped. Despite several commits trying to fix the issue, I'm still having that issue quite frequently.

Instead of calling Packery at the document's `DOMContentLoaded` event, or at jQuery's `$(document).ready` event, have you considered calling it at `$(window).on("load", ` event? This one guarantees that images have finished loading before calling packery, giving me consistent correct results, despite images taking a bit longer to go to their places, which I don't mind.

MaikelChan:
Another thing I noticed, since the ID of a gallery contains a Unix time, each time a site is built, each page that contains a gallery will change and will have to be committed to source control even if it technically has no changes, so the site generation is non-deterministic. Is that Unix time necessary?

PR #738, opened 2023-05-23
Gallery fixes.

MaikelChan:
- Removed `now.UnixNano` time from the IDs of the galleries so page generation is deterministic. This fixes generated pages with galleries always having differences after each build, even if the page hasn't changed. With `.Ordinal`, the galleries IDs should already be unique.
- Make sure that all images have finished loading before calling Packery, fixing some image overlapping problems.


nunocoracao:
Thanks @MaikelChan really appreciate this one

PR #748, opened 2023-05-28
Fix .Language.Params deprecation

mbrossard:
Since Hugo v0.112, `.Language.Params` has been deprecated in favor of `.Site.Params`.  Running with `--panicOnWarning` displays the following warning:
```
.Language.Params is deprecated and will be removed in a future release. .Language.Params
is deprecated and will be removed in a future release. Use site.Params instead.

- For all but custom parameters, you need to use the built in Hugo variables, e.g. site.Title, 
  site.LanguageCode; site.Language.Params.Title will not work.
- All custom parameters needs to be placed below params, e.g. [languages.en.params] in TOML.

See https://gohugo.io/content-management/multilingual/#changes-in-hugo-01120
```

The consequence of this PR would be to raise the minimum version of Hugo to v0.112, so it is probably too early to merge it. It will become necessary once the deprecation turns into removal.

### DIFF
--- a/layouts/_default/index.json
+++ b/layouts/_default/index.json
@@ -2,7 +2,7 @@
 {{- range .Site.Pages -}}
   {{- $section := .Site.GetPage "section" .Section -}}
   {{- $index = $index | append (dict
-    "date" (.Date | time.Format (.Site.Language.Params.dateFormat | default ":date_long")) 
+    "date" (.Date | time.Format (.Site.Params.dateFormat | default ":date_long")) 
     "title" (.Title | emojify | safeJS)
     "section" ($section.Title | emojify | safeJS)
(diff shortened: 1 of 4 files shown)


nunocoracao:
@mbrossard thanks for this. I need to deep dive a little into it and I really appreciate the help. 

A couple of questions/comments:
1) I think there are more changes than the code in the PR
2) documentation and config files need to be updated to reflect the changes  
3) can you explain why the need to raise the min version - though those params were supported in older versions

nunocoracao:
@mbrossard closing this one in favor of this PR https://github.com/nunocoracao/blowfish/pull/763 - just changed the config files which seems to solve it. Also easier to manager without increasing the minimum version - not sure about this one.
Happy to work on that one if you want to help.

PR #753, opened 2023-06-04
increased max version to Hugo latest v0.112.7

nunocoracao:


nunocoracao:
fixed https://github.com/nunocoracao/blowfish/issues/752

PR #773, opened 2023-06-10
⚙️ Remove and ignore .DS_Store files

michaeldorner:
I found some macOS-specific files in the project. I removed them and add those files to `.gitignore`. I also added
```
._*
.Spotlight-V100
.Trashes
```

to `.gitignore` which I think simply make sense.


nunocoracao:
Thanks @michaeldorner

PR #774, opened 2023-06-10
 Add hugo-easy-gallery support

sizeofvoid:
- Add and enable hugo-easy-gallery support from https://github.com/Darthagnon/hugo-easy-gallery/releases/tag/1.1

- Infos:
 - https://github.com/Darthagnon/hugo-easy-gallery/blob/master/README.md


ashcorpdev:
This project doesn't seem to be maintained anymore and also seems to have a few issues, such as [not supporting loading images from page bundles](https://github.com/Darthagnon/hugo-easy-gallery/issues/6#issuecomment-1505862821). The included gallery shortcode seems more reliable and supports page bundles by default.

nunocoracao:
@sizeofvoid please open a ticket to discuss topics like this before opening a PR. As @ashcorpdev mentioned the project in question is not updated for more than a year so I don't think we should add support for this in the theme. It is still possible for users that prefer to include it in their projects without it being included in the theme. Sorry if you invested time in this PR for me to close it now, let's discuss beforehand big changes like this one.

PR #779, opened 2023-06-15
🐛 fix regression in fathom analytics

voigt:
With the latest refactor, the site id for fathom analytics is not correctly set. This PR should fix it.


nunocoracao:
Thanks @voigt

Issue #798, opened 2023-07-05
⚙️ Support for Hugo v0.115.1 


nunocoracao:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #799, opened 2023-07-05
⚙️ Support for Hugo v0.115.1

nunocoracao:


nunocoracao:
fixes https://github.com/nunocoracao/blowfish/issues/798

PR #805, opened 2023-07-08
🎏 fix German translation of related_articles

fuse314:


nunocoracao:
Thanks @fuse314 will review for next release

PR #811, opened 2023-07-12
🐛 Fix for wrong Copyright in RSS feeds

MaikelChan:
- Copyright in RSS feeds would show the "{ year }" code instead of the proper year. Also, the logic matches the one in the footer of the page.
- Changed the language code to match with the [current Hugo default RSS template](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_default/rss.xml).
- Renamed index.rss.xml to rss.xml. Previously, the template provided with blowfish would only load in the root of the site, all other sections would load the default Hugo template, which has the Copyright issue. Now the provided template loads properly everywhere in the site. Unless it was intentional that it only loads in the root?


nunocoracao:
Thanks @MaikelChan

PR #814, opened 2023-07-14
🎨 Fix that prevents the TOC from being covered by the main menu.

MaikelChan:
The TOC can be partially hidden by the main menu when scrolling down if that menu has a logo image. That's even visible in the Blowfish page. Just to test that it wasn't a Windows thing, I also checked on Ubuntu and same result.

I adjusted the top value so the TOC never goes that high in the page when scrolling down.

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -6871,7 +6871,7 @@ body:has(#menu-controller:checked) {
   }
 
   .lg\:top-10 {
-    top: 2.5rem;
+    top: 6rem;
   }
 
   .lg\:top-\[140px\] {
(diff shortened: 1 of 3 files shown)


MaikelChan:
I actually found the proper fix.

It turns out there was a space in the " fixed" header layout type, so that check never returned true. That would set the `lg:top-10` class when it should be `lg:top-[140px]` for fixed header layout. I also changed the condition to check if the layout type starts with "fixed", as the `lg:top-[140px]` class should be applied to all fixed layout variants (fixed, fixed-fill, fixed-fill-blur).

Thus, I reverted the CSS change.

nunocoracao:
Thanks @MaikelChan

PR #815, opened 2023-07-14
🎏 Update to the Spanish translation

MaikelChan:
Added missing texts and some corrections:

- More consistent capitalization.
- Avoid using the word "Link" which is an anglicism.


nunocoracao:
Thanks @MaikelChan

PR #830, opened 2023-07-17
Always generate appearance.js in js folder

fuse314:
depending on the basePath, the "appearance.js" was generated in an additional subfolder (Example: set the basePath to "https://example.com/foobar/", the file was generated into "/js/foobar/appearance...js" instead of "/js/appearance...js" ).

This fixes the behaviour, the javascript resource is always generated into the target "js" folder along with all the other js files.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -50,7 +50,7 @@
   <link type="text/css" rel="stylesheet" href="{{ $bundleCSS.RelPermalink }}"
     integrity="{{ $bundleCSS.Data.Integrity }}" />
   {{ $jsAppearance := resources.Get "js/appearance.js" }}
-  {{ $jsAppearance = $jsAppearance | resources.ExecuteAsTemplate $jsAppearance.RelPermalink . | resources.Minify | resources.Fingerprint "sha512" }}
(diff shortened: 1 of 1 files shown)


nunocoracao:
Not sure about this one @fuse314. What is this fixing?

fuse314:
Steps to reproduce:
1. open config.toml, set ``baseURL = 'https://example.com/foo/bar/baz/'``
2. delete local "public" directory
3. run ``hugo``
4. check content of "public" directory for a directory called "foo":

example:
image

the pull request always generates the javascript file into the "js" directory, along with the other javascript files.

nunocoracao:
@fuse314 which given that you defined a baseUrl like that makes sense right? What exactly doesn't work when you do this?

fuse314:
@nunocoracao no, it does not. If I deploy the ``public`` folder to ``/foo/bar/baz/`` on the webserver, the full url to this file is ``/foo/bar/baz/foo/bar/baz/js/appearance...``.
It is the only file in the double relative path, all other js files are deployed in the ``/js/`` folder.
Just try it out locally and look at the resulting folder structure and the files.

I have various existing folders in my webroot and I have to create "include" and "ignore" rules for ftp deployment. It just makes these rules unneccessarily more complex.

PR #861, opened 2023-07-30
🎨 The disableImageOptimization setting now affects author images

MaikelChan:
The disableImageOptimization setting now affects author images, like the rest of the images.


nunocoracao:
Thanks for this one @MaikelChan

PR #893, opened 2023-08-07
📖 various docs improvements

jamesbraza:
- Addresses `markup.toml` confusion from https://github.com/nunocoracao/blowfish/issues/882 (I totally missed that the theme has a defaulted `markup.toml`)
- Removes unused `article.showSummary = true` from `params.toml` per [this discussion](https://github.com/nunocoracao/blowfish/discussions/892)
- Documents `hugo.toml` vs `config.toml`
- Removes duplicate "supports" word

PR #924, opened 2023-08-18
🔧 Removed `article.showSummary`

jamesbraza:
`list.showSummary` is used, `article.showSummary` is unused.  This PR removes all `article.showSummary` instances from the site.

Follow up to https://github.com/nunocoracao/blowfish/pull/893.


nunocoracao:
Thanks for this one @jamesbraza

PR #925, opened 2023-08-18
📖 Documenting support of HTML table

jamesbraza:
- Adds example HTML table to `markdown` sample
- Expands installation instructions for copying basic config files, as `markup.toml` items are required for HTML table support

Follow up to https://github.com/nunocoracao/blowfish/pull/893.


nunocoracao:
Sorry @jamesbraza closing this one. This is redundant and not required. HTML tags within the markdown are supported by Hugo not Blowfish

PR #975, opened 2023-09-28
🎏 Adding arabic languag + shortcodes for ltr/rtl

ajmandourah:
Arabic was not added as a language . I have seen an old PR but for some reason was not implemented. 
I added my translations that i have been using for my personal site. 

Additionally, mixing ltr/rtl content is something many need especially in a blog post . I added 2 shortcodes where you can implement in the post where the content in between will be rendered as the desired direction 

``` markdown
{{% rtl %}}
This content will be rendered as RTL
- this also
- this also
{{% /rtl %}}
```


nunocoracao:
@ajmandourah thanks for this
A couple of requests:
1. Add the two shortcodes to the documentation so that people know how to use them - please include an example
2. Could the shortcodes leverage tailwind and the classes already created for ltr and rtl rendering - wondering if in the future we change something the Shortcodes should follow the same logic without requiring separate code fixes.

ajmandourah:
- will work on the docs this weekend.
- makes sense regarding tailwind. Alas my experiece in it is minimal. The purpose for now of the shortcodes is when mixing contents in posts is desired. where most of the converted markdown have no css classes or ids.

ajmandourah:
Docs updated.
for the TW class, the text-aligment is probably not needed, its more of an assurance that the contents alignment will match the direction of the language. nevertheless, I included it in a TW-ready specified class.

nunocoracao:
Thanks @ajmandourah will try to merge to dev this weekend

PR #1001, opened 2023-10-11
add vscode code-snippets for blowfish shortcodes

wolfspyre:
Not sure if you want these or not; but thought I'd share.

Here are some pseudo-opinionated code-snippets for vscode. they make using the shortcodes in blowfish a bit easier.

Alongside the shortcodes are Markdown versions of the vscode code-snippets documentation, and the textmate code-snippets documentation that vscode's implementation attempts to be compatible with.

One of the inherent downsides of the tool is there is not an 'optional' tabstop.
one is given two options with tab-iteratable parameters: 
- tab-stop-with-free-text
- tab-stop-with-a-set-list-of-values, values may not be null.

This makes it a little obtuse to enumerate all possible shortcode parameters for ease of recall.

My solution is to create a static list tabstop for optional parameters, with the options of
- a space
- a space, plus the optional parameter and a generic value.

This allows the user to tab-through all available parameters and drop in the ones they wish to use, and changeup the values with as little hassle as possible.

Where it made sense to me, the code snippet will populate the shortcode `inner` with:
- the text that was/is highlighted when the code-snippet was invoked
  - if no text was highlighted, the contents of the clipboard will be added to the `inner` content

These may be invoked from within vscode by typing 
- `BFS-` (BlowFish Shortcode) 
- `HSC-` (Hugo ShortCode)  
  - I have several other code snippets files for different contexts. I have another file which contains all the native hugo shortcodes. Those are also matched by this pattern.
- `the shortcode's name`

This allows one to pull up a contextual list fairly easy of 
BFS- (present a list to me consisting of all BlowFish Shortcodes)
HSC- (present a list to me consisting of all shortcodes relevant to hugo)

I'm not sure this philosophy works for everyone. 

but I figured it wouldn't hurt to share in case it works for some.

### DIFF
--- a/.vscode/blowfish.code-snippets
+++ b/.vscode/blowfish.code-snippets
@@ -1,0 +1,211 @@
+{
+  "alert": {
+    "prefix": ["BFS-alert", "HSC-alert", "alert"],
+    "body": [
+      "{{< alert ${1| ,icon=\"iconName\"|}${2| , iconColor=\"iconColor\"|}${3| , cardColor=\"cardColor\"|}${4| , textColor=\"textColor\"|} >}}",
+      "${TM_SELECTED_TEXT:${CLIPBOARD}}$0",
+      "{{< /alert >}}"
+    ],
+    "description": "Outputs its contents as a stylised message box within your article. Blowfish alert Shortcode.  Documentation: https://blowfish.page/docs/shortcodes/#alert ",
+  },
+  "article": {
+    "prefix": ["BFS-article", "HSC-article", "article"],
+    "body": [
+      "{{< article link=\"${1:RelPermalink_Of_Target_Article}\">}}",
+      "$0"
+    ],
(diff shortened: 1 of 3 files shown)


nunocoracao:
Thanks @wolfspyre this looks really interesting. Let's merge and see what people think.

nunocoracao:
@wolfspyre how would I leverage this snippets as a user? - was not able to get the contextual list

wolfspyre:
in vscode, enable snippets, and add them to a project (or globally): 

snippets-configure
snipptets-in-use
snippetsEnablement

wolfspyre:
once the snippets are enabled you can configure them as specific to a language/filetype, or globally ... I just enabled them globally because I was trying to figure out how they work... I'm not sure if vscode's snippets is where I'll stay with these, or if implementing them in something like  alfred will be a better long term.... 

¯\_(ツ)_/¯ 

really I just wanted a reminder of all the various options and possible values, as I have the memory of a goldfish ;)

PR #1015, opened 2023-10-24
feat: exclude external pages from Sitemap

fuse314:
- Include pages with relative externalUrl with correct URL in Sitemap
- Exclude pages with absolute externalUrl from Sitemap
- Include pages with relative externalUrl with correct URL in Sitemap
- Improve output XML (no more linebreaks and spaces)


fuse314:
The isset check for externalUrl did not work correctly before, pages with relativeUrl were still included in the Sitemap XML.
[see sitemap.xml of blowfish.page](https://blowfish.page/sitemap.xml) and search for `https://blowfish.page/users/`

nunocoracao:
Thanks for this @fuse314

PR #1024, opened 2023-10-26
Provide source and attribution for zoom.min.js

TheDidek:
Add attribution for origin repository. This allows source check and easier modifying of zoom behavior without grepping GitHub guessing what library is used.


nunocoracao:
@TheDidek why include the non-minimized version of the JS lib?

nunocoracao:
Want to close a release today @TheDidek will include this PR but will remove the source. Let's discuss here or open a new ticket for that addition. Thanks

TheDidek:
It is a practice to prevent bad scenarios when linked repos disappear.
Also it make source available to the visitor up in front, on many projects user can remove .min and get the source.
But that is more marginal case, the important part was to add the attribution :).

nunocoracao:
@TheDidek did you test if including the source in the static folder exports it to the final site? I would assume so

PR #1033, opened 2023-11-01
Updating Twitter Icon

moccolfc:
Hi Nuno,

What are your thoughts on updating the Twitter icon to the new "X" icon?


nunocoracao:
@moccolfc thanks I'll do it. This SVG is not properly formatted e.g. missing the path param.

moccolfc:
Ah apologies. Thanks for sorting!

PR #1063, opened 2023-11-21
Feature/hero-image gradient parameter

law:
This adds a boolean site-parameter to control the application of the `bg-gradient-to-r` darkening css on a Hero image.  For backwards compatibility/'principle-of-least-astonishment', this parameter is set to 'true' by default, maintaining the previous default behavior of darkening the hero image via CSS.  

This also applies the Tailwind CSS 'drop-shadow-md' property to the Author and Site Title sections.  This will be invisible over a dark image, but aids in readability over a light image (such as what may get displayed if homepageImageGradientOverlay is set to false).  

Please don't hesitate to suggest changes if there is a cleaner/better/less-hateful way to do this - I am not the strongest when it comes to FE HTML/CSS stuff, but I am always eager to learn and improve.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -2,4 +2,4 @@
   [module.hugoVersion]
     extended = true
     min = "0.87.0"
-    max = "0.120.3"
+    max = "0.120.4"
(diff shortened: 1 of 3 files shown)


nunocoracao:
Hello @law thanks for the PR. I am sorry but I'll be closing it whatever. Please open a ticket so that we can discuss if this is the best implementation. E.g. this could be a different type of background. Also, we would need to add the relevant documentation so that people can use it.

PR #1070, opened 2023-11-26
Update users.json

warnotte:
I tried that pull request


GeorgeGedox:
I don't think this is going to be merged as is mate.
You need to add your website at the end of the list and remove the `Theme author` tag, that's reserved for @nunocoracao

nunocoracao:
Thanks @warnotte I'll close the PR and manually add the user to the list

PR #1071, opened 2023-11-28
Only send web request for zoom library if it is not disabled.

ItsJustRuby:
There is already [an if condition](https://github.com/nunocoracao/blowfish/blob/45d634f6ca698abdbfebf67e955211f090043e5b/layouts/partials/footer.html#L46) in `footer.html` that only runs the JS from the zoom library if it is enabled per-site.

This commit takes the same if condition and applies it in `head.html` as well to save a unnecessary web request to `zoom.min.js` if that is the case.


nunocoracao:
looks good thanks @ItsJustRuby

Issue #956, opened 2023-09-17
Missing pointer-cursor for tag-badge in article view


JVimes:
**Describe the bug**

The pointer-cursor doesn't work for tag-badges in articles (works in cards though).

![image](https://github.com/nunocoracao/blowfish/assets/6148201/c06f547a-a246-4597-ba4f-7cd901afe6ee)

PR #1078, opened 2023-12-01
Missing pointer-cursor for tag-badge in article view

nunocoracao:
Fixes https://github.com/nunocoracao/blowfish/issues/956

### DIFF
--- a/layouts/partials/article-meta/basic.html
+++ b/layouts/partials/article-meta/basic.html
@@ -47,7 +47,7 @@
 {{ end }}
 
 
-<div class="flex flex-row flex-wrap items-center">
+<div style="cursor: default;" class="flex flex-row flex-wrap items-center">
   {{/* Output partials */}}
   {{ with ($meta.Get "partials") }}
   {{ delimit . "<span class=\"px-2 text-primary-500\">&middot;</span>" | safeHTML }}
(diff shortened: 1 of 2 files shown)


layouts/partials/article-meta/basic.html
```diff
@@ -47,7 +47,7 @@
 {{ end }}
 
 
-<div class="flex flex-row flex-wrap items-center">
+<div style="cursor: default;" class="flex flex-row flex-wrap items-center">
```
JVimes, line 50: @nunocoracao I think this hides the pointer cursor for article date link. Is that on purpose?

nunocoracao: @JVimes yes, did you have any use for it?

JVimes: Without it, I'd assume the site doesn't link article dates.

nunocoracao: Sorry @JVimes I am not following the use-case. Can you point me to an example of what worked and doesn't anymore. Maybe we should open a issue for this also.

JVimes: Never mind, I'm getting my wires crossed with my modifications 🙂

Issue #1125, opened 2023-12-18
🐛 Shortcodes Gallery : Images are not correctly displayed


v20100v:
**Describe the bug**

When you display only one or more galleries on the page, the images are too small and do not take up the available space.

![image](https://github.com/nunocoracao/blowfish/assets/7567933/78a3eaee-9980-4e3f-a300-38b819f13b3b)

But if we add enough text in the column, then the images are correctly displayed... unless the text takes up too much space and in this case the images are no longer correctly aligned horizontally.

![image](https://github.com/nunocoracao/blowfish/assets/7567933/327e1ba0-3e02-42c0-8f49-6c2d61cc16f9)

**To Reproduce**

1. Create new blowfish project 

2. Create this post
```html
---
title: "Hugo theme Blowfish - Bug gallery"
date: 2023-12-15
---

### First

{{< gallery >}}
  <img src="amstrad-464/img_1.png" class="grid-w50 md:grid-w33 xl:grid-w25" />
  <img src="amstrad-464/img_2.png" class="grid-w50 md:grid-w33 xl:grid-w25" />
  <img src="amstrad-464/img_3.png" class="grid-w50 md:grid-w33 xl:grid-w25" />
  <img src="amstrad-464/img_4.png" class="grid-w50 md:grid-w33 xl:grid-w25" />
{{< /gallery >}}

### Second

Lorem ipsum.

{{< gallery >}}
<img src="amstrad-464/img_4.png" class="grid-w50" />
<img src="amstrad-464/img_5.png" class="grid-w50" />
{{< /gallery >}}
```

**Expected behavior**

Let the gallery(s) occupy the entire width.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser : Edge
 - Version : 120.0.2210.61 
 
**Hugo & Blowfish versions**
Which version of Hugo and Blowfish are you using?

- hugo v0.120.4-f11bca5fec2ebb3a02727fb2a5cfb08da96fd9df+extended windows/amd64 BuildDate=2023-11-08T11:18:07Z VendorInfo=gohugoio
- Blowfish : 2.47.2

v20100v:
Looking at the code, I noticed another problem.

When displaying several galleries on the same page, they may share the same ID. This occurs when galleries are included in a `timelineitem`.

```html
---
title: "Hugo theme Blowfish - Bug galleries in timeline"
date: 2023-12-15
---


### Galleries in timeline

{{< timeline >}}

[...]

PR #1126, opened 2023-12-18
🐛 Shortcodes gallery (non-unique id, multiple loading <script> tag)

v20100v:
Related to bug issue https://github.com/nunocoracao/blowfish/issues/1125

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ node_modules
 ._*
 .Spotlight-V100
 .Trashes
+.idea
--- a/assets/js/shortcodes/gallery.js
+++ b/assets/js/shortcodes/gallery.js
@@ -1,0 +1,24 @@
+function _getDefaultPackeryOptions() {
+    return {
+        percentPosition: true,
+        gutter: 5,
+        resize: true
+    };
+}
+
+(function init() {
+    window.addEventListener("DOMContentLoaded", (event) => {
+        console.groupCollapsed('[DEBUG] Gallery feature enable');
+        let packeries = [];
+        let nodeGalleries = document.querySelectorAll('.gallery');
+
+        nodeGalleries.forEach(nodeGallery => {
+            // TODO : implement a reader of Packery configuration _getPackeryOptions; for example by reading data-attribute
+            let packery = new Packery(nodeGallery, _getDefaultPackeryOptions());
+            packeries.push(packery);
+        });
+
+        console.log("Galleries founded and initialized with packery", packeries);
+        console.groupEnd();
+    });
+})();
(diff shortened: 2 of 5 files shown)


v20100v:
Some proposals in this merge request

### Changes in ./shortcodes/gallery.html

1. Proposition for a UUID more robust 
   ```
   {{ $random := delimit (shuffle (seq 1 9)) "" }}
   {{ $id := delimit (slice "gallery" $random now.UnixNano) "-" }}
   ```
2. Remove the tag <script>, because it was included multiples times when use multiples shortcode gallery in markdown. We need to include once. To do it, we create a specific js (./gallery.js) included in ./layouts/partials/vendor.html


### Changes in package.json

Other proposals (which have nothing to do with the shortcode gallery)

1. When you first deploy this project in a dev environment, it's strange that you have already installed the node package Rimraf as global dependencies (otherwise you get an error: package not installed). This package is install as a local node development dependencies. Remove preinstall ?

2. The custom node script `fullinstall": "npm run preinstall && npm install && npm run postinstall"` is not necessary. When you use `npm install`, the `preinstall` and `postinstall` node scripts are run automatically by npm before and after the package is installed, respectively. 

3. Add scripts (dev-windows & build-windows) compatibles on Windows system (see cross-platform env variables in Node.js)

MaikelChan:
Turns out that this PR has brought back the issue of overlapping images in the gallery.

v20100v:
Hello @MaikelChan,

I'm looking in detail, and trying to find a solution on this overlapping image problem.You have undoubtedly found the right clue, on the problem of loading the complete DOM with the Packery library.

nisetynet:
`blowfish/assets/js/shortcodes/gallery.js`

@v20100v 
Hello.
Thank you for developing the blowfish theme.

Based on MaikelChan's observations about the problem of overlapping images, what about the following code?
I have tested it dozens of times, and the problem does not occur.
At least in my environment.

```js
function _getDefaultPackeryOptions() {
    return {
        percentPosition: true,
        gutter: 5,
        resize: true
    };
}

(function init() {
    console.debug("Set on load callback.");

    $(window).on("load", function () {
        console.groupCollapsed('[DEBUG] Gallery feature enable');
        let packeries = [];
        let nodeGalleries = document.querySelectorAll('.gallery');

        nodeGalleries.forEach(nodeGallery => {
            // TODO : implement a reader of Packery configuration _getPackeryOptions; for example by reading data-attribute
            let packery = new Packery(nodeGallery, _getDefaultPackeryOptions());
            packeries.push(packery);
        });

        console.log("Galleries founded and initialized with packery", packeries);
        console.groupEnd();
    });
})();
```

Issue #994, opened 2023-10-09
Article lists are incredibly laggy on mobile Firefox and overall theme performance is very low


ragibson:
**Describe the bug**
Scrolling through a list of articles is very laggy and jagged on Firefox. Overall performance of the theme is also slow.

**To Reproduce**
This appears to occur on pretty much every Blowfish website, including https://blowfish.page/docs/.

EDIT: For example,
* The Blowfish documentation [achieves 56/100 on mobile PageSpeed Insights](https://pagespeed.web.dev/analysis/https-blowfish-page-docs/iyzeugbsgu?form_factor=mobile) whereas most popular Hugo themes achieve at or near 100%.
* The ["Welcome to Blowfish" article hits 35% on mobile and 72% on desktop](https://pagespeed.web.dev/analysis/https-blowfish-page-docs-welcome/mrebs8zx2k?form_factor=mobile).

**Expected behavior**
The page scrolls smoothly and overall performance is improved.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Android
 - Browser: Firefox
 - Version: 118.1.1 and has persisted for many months beforehand

**Hugo & Blowfish versions**
The scrolling issue is on most sites I have checked and the issue has persisted for months, so it should affect all recent versions.

**Additional context**
N/A

ragibson:
Importantly, this issue does not appear to be present in the upstream [Congo](https://github.com/jpanther/congo).

LITUATUI:
Seems fine on Linux using the version 118.0.2 (64-bit).

ragibson:
I'll edit the issue title back to the original -- this is predominantly a mobile effect

ragibson:
Here, the Blowfish documentation [achieves 56/100 on mobile PageSpeed Insights](https://pagespeed.web.dev/analysis/https-blowfish-page-docs/iyzeugbsgu?form_factor=mobile) whereas most popular Hugo themes achieve at or near 100%.

ragibson:
Actually, even the ["Welcome to Blowfish" article hits 35% on mobile and 72% on desktop](https://pagespeed.web.dev/analysis/https-blowfish-page-docs-welcome/mrebs8zx2k?form_factor=mobile), so this is likely a device agnostic problem.

jamesbraza:
A related issue previously observed is use of `layoutBackgroundBlur` lead to slowness: https://github.com/nunocoracao/blowfish/issues/907

[...]

Issue #907, opened 2023-08-12
🐛 `layoutBackgroundBlur` + `showRecent` leads to laggy homepage


jamesbraza:
**Describe the bug**

When working on my portfolio website, I observed homepage settings `layoutBackgroundBlur = true` with `showRecent = true` and `showRecentItems = 12` lead to a laggy homepage when scrolling down. I disabled `layoutBackgroundBlur`, and the issue went away.

Looking at [this site](https://bbagwang.com/) ([config](https://github.com/bbagwang/bbagwang.github.io/blob/main/config/_default/params.toml)), it seems to have the same issue (as of 8/12/2023).

**To Reproduce**

1. Set `layoutBackgroundBlur = true` and `showRecent = true` (with `showRecentItems` >= 10) in `[homepage]` section of `params.toml`
2. Go to homepage and scroll down

The lag happens for both a Namecheap shared hosting (what I use) and GitHub Pages (linked site).

**Desktop (please complete the following information):**

 - OS: macOS Monterey version 12.6
 - Browser: Google Chrome 115.0.5790.170
 - Site: Hugo 0.116.1, Blowfish 2.38.3

Here's a [Speedtest](https://www.speedtest.net/) result, to show I have decent internet speeds:

<img width="362" alt="screenshot of speedtest" src="https://github.com/nunocoracao/blowfish/assets/8990777/7754945d-9870-4bb7-a03e-66a0d9ed7349">

ragibson:
Popping in here as well from the other issue to note that I found a lot of documentation saying to avoid these blur filters entirely as they are extremely intensive and often do not perform well on mobile.

The pull request linked above improves the draw calls in Firefox by ~2x when using the blur features, but a remaining ~100x would come from a better CSS implementation that avoids `backdrop-filter` entirely.

nunocoracao:
Should be fixed. Please re-open if the issue persists

PR #1128, opened 2023-12-18
Remove extraneous backdrop-blurs across partials

ragibson:
This should mostly fix https://github.com/nunocoracao/blowfish/issues/994 and partially address https://github.com/nunocoracao/blowfish/issues/907.
- Removes most of the extraneous usage of `backdrop-blur` with no apparent visual impact (I only kept it on menu-wrapper, menu-blur, nested-menu, background-blur, and search-wrapper)
- Improves number of draw calls in Firefox with `header = "fixed"` and `layoutBackgroundBlur = false` by **~200 times**, drastically improving scrolling performance
- Improves number of draw calls in Firefox with `header = "basic"` and `layoutBackgroundBlur = true` by ~50%

In my earlier testing, I had simply disabled `.backdrop-blur`, `.backdrop-blur-sm`, and `.backdrop-blur-2xl` on every element except for the search widget, but this is a more proper solution. The inclusion of the `backdrop-filter` on every article link was incredibly intensive and seemingly included as an oversight as it doesn't appear to have any visual effect. That said, please review and check if any of them were actually intended.

Probably more could be done with a better CSS implementation since I did find a lot of documentation saying to avoid these blur filters entirely as they are extremely intensive and often do not perform well on mobile.

The draw call statistics above are from a local build of your website, https://github.com/nunocoracao/n9o.xyz.
n90_comparison

### DIFF
--- a/layouts/partials/article-link/card-related.html
+++ b/layouts/partials/article-link/card-related.html
@@ -4,7 +4,7 @@
   {{ else }}
   <a href="{{ .RelPermalink }}" class="min-w-full">
     {{ end }}
-    <div class="min-h-full border border-neutral-200 dark:border-neutral-700 border-2 rounded overflow-hidden shadow-2xl relative backdrop-blur">
+    <div class="min-h-full border border-neutral-200 dark:border-neutral-700 border-2 rounded overflow-hidden shadow-2xl relative">
 
       {{- with $.Params.images -}}
       {{- range first 6 . }}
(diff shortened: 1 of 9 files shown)


nunocoracao:
Thanks for this @ragibson will need time to review. I will probably merge the PR and make some extra changes. Did you test other setups - aka a config of the theme that doesn't use background-blur?
Those background blurs on the partials were meant for people not using the background blur on the page itself. Might need to revisit this.

ragibson:
Yes, my site has the background blurs disabled and it didn't seem to have
any real visual impact (though they definitely have a performance impact).

Just wanted to check on the purpose of them on some of the elements that
don't *seem* to be visually affected by blur at all. It was included on
quite a few.

nunocoracao:
Give me some time to test this.

PR #1136, opened 2023-12-20
:sparkles: Keyword shortcode

alxhslm:
This PR implements #1109, adding a keyword component. I'm a Python developer so all of this HTML is new to me, but I've got something that works. Feel free to make any changes or suggest any improvements.

Screenshot 2023-12-20 at 16 17 43


nunocoracao:
Thanks, @alxhslm will need some time to review this. I am pondering if we need a separate implementation or if we should extend the badge shortcode. Thoughts?

PR #1138, opened 2023-12-21
Update timelineItem.html

xwjdsh:
* Add `break-words` to prevent content overflow

---
before:
image

after:
image


nunocoracao:
Thanks @xwjdsh will review

PR #1183, opened 2024-01-19
Add Xing icon

arunoruto:
This resolves #1111

Issue #1111, opened 2023-12-10
Xing.com as a supported link in languages.toml


chromer030:
**Is your feature request related to a problem? Please describe.**
Xing is a popular German business and employment-focused social media platform, please add it as supported.

PR #1204, opened 2024-01-30
Add xing support for social links

arunoruto:
As far as I know, the changes proposed in #1183 have been implemented and should solve #1111. I tested it on a local run, and the color was applied appropriately. The size is also the same as the other icons!


nunocoracao:
Still missing the doc entry but I'll add it

Issue #1211, opened 2024-02-04
[Feature] Option to load images from URLs


aakashnand:
**Is your feature request related to a problem? Please describe.**
Currently whenever I find documentation around images for this theme. It expects images to be added to the git repository. That is the image should be present in the `assets` folder and for the feature image, it should be present in each blog folder with a name like `feature**`. 

While this is great, I think this is not good as it will bloat the git repository. 

**Describe the solution you'd like**

Every image reference in the theme should have the option to display an image from some external URL or local reference. Giving both options to users will be a better experience and can reduce the repository size. 

**Describe alternatives you've considered**
I am not a Hugo expert but after some digging, I think it can be achieved `resources.GetRemote`using https://gohugo.io/functions/resources/getremote/

Although I have limited knowledge of how to implement this.

**Additional context**

Similar requests and discussions I found on this repo

1. https://github.com/nunocoracao/blowfish/discussions/880
2. https://github.com/nunocoracao/blowfish/discussions/932
3. #980 
4. #736

nunocoracao:
@aakashnand your PR doesn't solve #736 right?

aakashnand:
> @aakashnand your PR doesn't solve #736 right?

No it doesn't but this PR can be a good base for that issue as well. I think for carousel we might need to tweak around short-codes. 

If you want we can tackle that in different PR

nunocoracao:
fixed in https://github.com/nunocoracao/blowfish/pull/1216

PR #1216, opened 2024-02-05
Add ability to set feature image as URL

### LINKED ISSUE #1215: ❓ Featured images for different languages or in frontmatter
Hi! It works great when you create file index.en.md with post's text - it shows for specified language. But I haven't found any way to use different featured image for different languages.

By default use `*feature*` image name mask, as usual. If there is `.en.` or `.<langcode>.` in the file name - use it only for specified language. 

As alternative you should be able to set featured image path in a frontmatter, e.g. `featured_image: different_image.jpg`

✌️

aakashnand:
Thank you for this awesome theme @nunocoracao. 

When I started using this theme I faced a problem with setting images in my theme. The current version only supported images from the `asset` folder. This can make the git repository and history bigger if we want to use multiple images for each blog post. Many users already had raised issues about this feature. I summarized all of them in #1211 

To fix this, I have added the following functionalities. 

1. `homepage.homepageImage` can now be set as an external image URL. The theme will check if the file is present locally if not it will check if the file is present in the URL specified. 
2. Introduced new front matter configuration `featureimage`. This will be the external URL of an image to be used as a feature image.
3. Introduced new front matter configuration `featureimagecaption`. This will be the caption for the image which will be displayed in `article.heroStyle=big`
4. Updated documentation for frontmatter and configuration.

Let me know your feedback and if anything needs to be improved. 

Closes #1215


LITUATUI:
This is very useful if you're moving from a theme that supports feature image via URL.

aakashnand:
> Thanks @aakashnand I'll need some time to review this PR. My initial skim I would ask you to not do a style override in the single.html file. Either use the tailwind classes or create a new object in the main.css files that gets compiled into the final theme. Inline styling is really hard to manage. Also kudos on getting docs sorted already :)

@nunocoracao. Understood. I will update and see what can be done. I also forgot to add documentation about the new config that I introduced.  `.Params.maxWidth` which allows customization from frontmatter.

nunocoracao:
@aakashnand please remove the extra parameter .Params.maxWidth from this PR. This is getting a little bloated and should be separate things. Also I don't think we should have that parameter in frontmatter

aakashnand:
Yep I think it was more like personal requirement than a contribution. I will remove the maxWidth parameter today

aakashnand:
@nunocoracao removed maxWidth. Let me know what do you think now.

nunocoracao:
@aakashnand I think it's just missing a check for whether the URL is defined in frontmatter before trying the getRemote calls. Also (and I didn't have time to check everything yet) I was not sure if the logic is the same for all? i.e. local image -> remote -> site default

aakashnand:
> @aakashnand I think it's just missing a check for whether the URL is defined in frontmatter before trying the getRemote calls. Also (and I didn't have time to check everything yet) I was not sure if the logic is the same for all? i.e. local image -> remote -> site default
> 
> 

Let me recheck. If you can point your specific line that would be super helpful

aakashnand:
@nunocoracao 

I made the changes you suggested. 
1. Added a check to see if `featureimage` is present in the front matter before calling `GetRemote`
2. Also previously I forgot to add `GetRemote` for the author image. Now author image can be either present in the `asset` folder or can be set as an external URL as well. 
3. Formatted code to enhance readability. 
4. Added documentation for `author.image` saying that image can point to external URL now

nunocoracao:
sweet :) thanks a lot

nunocoracao:
@aakashnand I ended up making "some" changes https://github.com/nunocoracao/blowfish/pull/1213/commits/e87e066aa7250ac033ca01959639cbf795b14614

aakashnand:
@nunocoracao Thank you so much for making the additional changes. I am glad I found this theme. I will soon send PR to add my site as a user

PR #1233, opened 2024-02-11
Improve consistency in header's layout and size

MaikelChan:
Hi!

I just fixed some inconsistencies in the header:
- In mobile version, its height would depend on the height of the dark mode switcher. If the user has the switcher disabled, the header would appear very small on mobile.
- Dark mode switcher is smaller on desktop mode than on mobile, and compared to the other icons.
- Dark mode switcher on mobile doesn't have color change on hover, the other buttons have.
- Clickable area in dark mode switcher is much bigger than the other icons (that's why the height of the header in mobile depended on this button).

### DIFF
--- a/layouts/partials/header/basic.html
+++ b/layouts/partials/header/basic.html
@@ -35,7 +35,7 @@
             {{ partial "translations.html" . }}
 
             {{ if .Site.Params.enableSearch | default false }}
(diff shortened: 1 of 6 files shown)


MaikelChan:
I also added hover color to the header menu and nested menus, the same as the search and dark mode switcher buttons. They didn't have any reaction to being hovered.

PR #1235, opened 2024-02-11
Fix background image mode in list view

aakashnand:
Fixes comment from

> This line was added to ensure that we fetched the background image using an external URL even in list, term, and taxonomy mode when showHero was set to true.
> 
> Currently, this is causing a bug. When hero mode is set to true, in the list view, it's not showing the background image.
> 
> I think we should keep the if condition. This condition checks if hero mode is set to true then it tries to fetch the image from the externalURL

https://github.com/nunocoracao/blowfish/commit/e87e066aa7250ac033ca01959639cbf795b14614#r138507821

### DIFF
--- a/layouts/partials/hero/background.html
+++ b/layouts/partials/hero/background.html
@@ -4,12 +4,17 @@
 {{- $featured := $images.GetMatch "*background*" -}}
 {{- if not $featured }}{{ $featured = $images.GetMatch "*feature*" }}{{ end -}}
 {{- if not $featured }}{{ $featured = $images.GetMatch "{*cover*,*thumbnail*}" }}{{ end -}}
-{{ if and .Params.featureimage (not $featured) }}
-{{- $url:= .Params.featureimage -}}
-{{ $featured = resources.GetRemote $url }}
-{{ end }}
-{{- if not $featured }}{{ with .Site.Params.defaultBackgroundImage }}{{ $featured = resources.Get . }}{{ end }}{{ end -}}
 {{ $isParentList := eq (.Scratch.Get "scope") "list"  }}
+{{- if not $featured }}
+{{ with .Site.Params.defaultBackgroundImage }}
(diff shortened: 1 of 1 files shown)


layouts/partials/hero/background.html
```diff
@@ -4,12 +4,17 @@
...
+{{ $isParentList := eq (.Scratch.Get "scope") "list"  }}
+{{ $url:= default nil }}
+{{ if $isParentList }}
+{{ $url = default $.Site.Params.homepage.homepageImage $.Site.Params.defaultBackgroundImage }}
+{{ else }}
+{{ $url = .Params.featureimage }}
+{{ if not $url }} {{ $url = default $.Site.Params.homepage.homepageImage $.Site.Params.defaultBackgroundImage }} {{ end }}
 {{ end }}
+{{- if not $featured }}{{ $featured = resources.GetRemote $url }}{{ end -}}
 {{- if not $featured }}{{ with .Site.Params.defaultBackgroundImage }}{{ $featured = resources.Get . }}{{ end }}{{ end -}}
```
aakashnand, line 16: Here is what these lines mean in plain English.

If the scope is `list` (that is list view) then set the URL as `homepageImage` or `defaultBackgroundImage`

else when it is article view set URL as `featureimage` if featureimage is not present set article background image as `homepageImage` or `defaultBackgroundImage`

nunocoracao:
Ups sorry @aakashnand

aakashnand:
@nunocoracao let me retry with your logic and test it again

aakashnand:
@nunocoracao made changes according to your logic. I think this way it is now easy to understand and much readable

nunocoracao:
Hey @aakashnand thanks for the effort, the code still has problems. You removed the loading for the actual front-matter option. Also, that code doesn't need to be inside the list stuff in order to work for anything. I'll close this and make a PR myself.

PR #1251, opened 2024-02-16
:globe_with_meridians: Add Catalan translation

albertonl:
I added the Catalan translation for the default i18n strings. I didn't add the language to the list in the [Configuration page of the documentation](https://blowfish.page/docs/installation/), because it is unclear what flag should be used (if any). For that same reason, I chose to apply the `CA` tag in the YAML file. Any suggestions on how to resolve this would be appreciated.

### DIFF
--- a/exampleSite/content/docs/configuration/index.md
+++ b/exampleSite/content/docs/configuration/index.md
@@ -67,6 +67,7 @@ The theme currently supports the following languages by default:
 | 🇦🇪 Arabic                       | `ar`    |
 | 🇧🇬 Bulgarian                    | `bg`    |
 | 🇧🇩 Bengali                      | `bn`    |
+| 🇪🇸 Catalan                      | `ca`    |
 | 🇨🇿 Czech                        | `cs`    |
 | 🇩🇪 German                       | `de`    |
 | 🇬🇧 English                      | `en`    |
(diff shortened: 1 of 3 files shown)


nunocoracao:
Hello @albertonl thanks for adding Catalan. I really appreciate the work you've put into this. Can you please add it to the list of languages in this PR. You brought up a really good point, the iso codes for countries are mixed up with codes for languages. I would suggest adding the country flag followed the language written in text. Feel free to propose something different if the file is not indexed in that list users will not find out about it.

albertonl:
Hi @nunocoracao, the issue is that no flag emoji is suitable enough for Catalan. One option would be to use the Spanish flag, but it has already been used for Spanish, and I believe it is not representative enough either.

Another option would be to use the Andorran flag (Catalan is the official language of Andorra), but I also have mixed feelings about that.

nunocoracao:
Why not Spanish flag with text (Catalan) after?

albertonl:
It was mostly to avoid repetition, but if you're okay with it I'll add it straight away.

nunocoracao:
@albertonl I am good with whatever you propose and then we can chat :)
I understand the mix up between languages and countries, and I am in no position to make comments on that.

albertonl:
@nunocoracao I'm fine with this solution, and thanks for understanding! :)

kahakai:
@nunocoracao I suggest you read https://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/ about the topic and never ever use flags for representing languages again to avoid confusion and other issues (I felt this throughout my life by myself). Although there are some arguments in favour of doing so. However, that's a whole another story for discussion not for this PR. You can check other blogposts regarding this with examples of software that removed the flags in the last years.

nunocoracao:
Legacy topic @kahakai which I never fixed... thanks for sharing

nunocoracao:
Thanks @albertonl

PR #1287, opened 2024-03-05
Add target parameter to figure shortcode

graeme-a-stewart:
Add the `target` parameter to the `figure` shortcode. If defined then add this target property to the `href` of the figure, which allows the link to, e.g., open in a new tab.

Added description of `target` parameter to docs.


nunocoracao:
Thanks @graeme-a-stewart

PR #1296, opened 2024-03-07
Gallery: Switch Packery to MagicGrid 

StandingPadAnimations:
This switches the `gallery` shortcode to use [MagicGrid](https://github.com/e-oj/Magic-Grid) instead of Packery. The main benefit is size, with MagicGrid being 2.7k, whereas Packery is 32k. The results are mostly similar minus a few differences in image position.


nunocoracao:
Hello @StandingPadAnimations not sure about this one. Changing a dependency like this can be a headache to maintain. I do like the decrease in size and also the fact that it seems more recent when compared to packery. I'll need time to pull your branch and see the differences.

nunocoracao:
Reverting @StandingPadAnimations I don't like the final effect

Packery
Screenshot 2024-03-09 at 14 29 02

Magic-grid
Screenshot 2024-03-09 at 14 29 17

PR #1298, opened 2024-03-07
🐛  Fix styling of Mermaid charts

alxhslm:
I noticed that Mermaid charts were no longer being updated to match the site theme. This made them particularly hard to read in dark mode. 

This appears to be due to a regression introduced in this [commit](https://github.com/nunocoracao/blowfish/commit/7e41c9cac0e9249dfde6d85990878ee7a36b8418).

Here is a comparison of the result before and after this PR:
Before
After


nunocoracao:
Ups, thanks for fixing this @alxhslm

PR #1351, opened 2024-03-20
blowfish docs translate zh-CN

Weaxs:


Aphcity (commented):
太效率了哥们，so much efficient

Aphcity (commented):
我觉得i18n下面那两个yml，language给他保留着 zh-cn zh-tw 就行，因为和别的语言一样都是用统一的iso代码。

I think those two YML files in "i18n" folder should remain the same as it is now. Keep the isocode zh-CN and zh-TW as it in other languages.

But as for the isocode, @nunocoracao I also suggest to use full length code for all language as a standard. Since when English(EN) or german(DE) is combined with Chinese(zh-CN), especially when the language name in the hide-able menu is longger more than 6 bytes, it would cause a visual format issue like pictures below. So I think we should change them into a universal style.

image

image

Weaxs:
> 我觉得i18n下面那两个yml，language给他保留着 zh-cn zh-tw 就行，因为和别的语言一样都是用统一的iso代码。
> 
> I think those two YML files in "i18n" folder should remain the same as it is now. Keep the isocode zh-CN and zh-TW as it in other languages.
> 
> But as for the isocode, @nunocoracao I also suggest to use full length code for all language as a standard. Since when English(EN) or german(DE) is combined with Chinese(zh-CN), especially when the language name in the hide-able menu is longger more than 6 bytes, it would cause a visual format issue like pictures below. So I think we should change them into a universal style.
> 
> image
> 
> image

语言列表展示的是` .Language.Params.displayName`， 但选中的语言会使用`language` 配置的值，这块有点奇怪...所以改了这个。嗯...不过确实还是最好在 layouts 里改应该更好点。

The language listing was displayed by ` .Language.Params.displayName`, but when I selected a langauge, it will display `language` value in `i18n/[language-code].yaml.
Maybe it is better to update layouts instead of changing `i18n/[language-code].yaml ?

image

I will fix it later.

Weaxs:
I saw `layouts/partials/translations.html`, only this layout file use `"language"` in`i18n/[language-code].yaml`.
I thought maybe  `"language"` in `i18n/[language-code].yaml` is only display name, so translate `zh-CN` to `简体中文` in  `i18n/zh-CN.yaml` is reasonable.

看了下 `layouts/partials/translations.html`，这里`i18n/[language-code].yaml`里面的 `"language"` 应该仅仅是一个展示名字，感觉改成 `简体中文`  是合理的。

```
<span class="mr-1">
      {{ partial "icon.html" "language" }}
    </span>
```
@Aphcity 
image

PR #1353, opened 2024-03-20
Update zh-CN translations

Aphcity:
- Update zh-CN trans for Advanced Customisation
- Update zh-CN trans for index
- Update zh-CN trans for posts


Aphcity:
btw, @nunocoracao should we create a trello board for tracking the exampleSite translation status, to check clearly which page still remained not fully translated.

nunocoracao:
@Aphcity I wold rather track it here in github to have a single platform but we can sure

Issue #1349, opened 2024-03-20
Cannot built with Hugo v0.123


yuntan:
**Describe the bug**
My site cannot be built with Hugo v0.123 due to `{{ .File }}` in HTML templates is nil.

**To Reproduce**
1. Install Hugo v0.123.7
2. `hugo`

**Expected behavior**
My site is built.

**Screenshots**
<img width="1331" alt="image" src="https://github.com/nunocoracao/blowfish/assets/574575/72b64d96-1f87-4b32-8968-84bcea17bfa0">

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser NA
 - Version NA

**Hugo & Blowfish versions**
Hugo: v0.123.7
Blowfish: v2.61.1

**Additional context**
Replacing `{{ .File.Path }}` in `layouts/partials/meta/views.html` and `layouts/partials/meta/likes.html` with `{{ .Path }}` makes build runs.

Related: #1282

nunocoracao:
Please update to the latest version of Hugo. 
Both my site and Blowfish's main page use vies/likes and are building correctly locally and in GH actions.

Aphcity:
**To Reproduce**
Steps to reproduce the behavior:
1. install hugo 0.124.1
2. run 'hugo'

**Screenshots**
![image](https://github.com/nunocoracao/blowfish/assets/16728561/a8bbfae5-49e7-4ff7-b84b-e45371213c72)

**GitHub Actions:**
 - OS: Ubuntu
 - Version 22.04.4 LTS

**Hugo & Blowfish versions**
Hugo 0.124.1
Blowfish 2.62.0

**Additional context**
Should be fixed in [PR#1359](https://github.com/nunocoracao/blowfish/pull/1359)

PR #1359, opened 2024-03-21
🐛 Fix list.html render issue

Aphcity:
FIXED: execute of template failed: template: _default/list.html:23:19: executing "main" at <.File.Path>: error calling Path: runtime error: invalid memory address or nil pointer dereference


Aphcity:
Should close ISSUE #1349

PR #1362, opened 2024-03-22
✨ Add player parameters for youtubeLite shortcode

d4rklynk:
https://github.com/paulirish/lite-youtube-embed?tab=readme-ov-file#custom-player-parameters


nunocoracao:
Hello @d4rklynk PR looks good, it's missing the documentaiton update. Can you please add to this PR?

d4rklynk:
Done, let me know what you think

nunocoracao:
@d4rklynk looks better - could you add the detailed list of parameters in the docs or surface them as individual parameters at the top level?

d4rklynk:
Do you mean explaining all the parameters described in the Google docs or just list them?

nunocoracao:
At least point people to where to find the parameters (sorry if it's already there and I missed it) if we keep the params var.

d4rklynk:
Indeed, I set it to the last paragraph.

I can rephrase though if you want.

nunocoracao:
Nothing need thanks. Sorry for missing it.

PR #1447, opened 2024-04-30
🎏 New gl.yaml for Galician language

jmmarino2:


nunocoracao:
Thanks @jmmarino2 please update the docs to mention this in the i18n section

PR #1452, opened 2024-05-04
Update config.toml

erdembey:
Bump Hugo


nunocoracao:
Thanks @erdembey the repo has an action that checks this and opens an automatic PR

PR #1458, opened 2024-05-08
Use AutoOrient filter when resizing images

karosas:
> Note - Haven't really developed for hugo, unsure if this has some unexpected consequences.

## What

Hugo, when resizing images, ignores EXIF data and suddenly some images (e.g. photos taken with a phone in portrait mode, I believe) end up rotated by 90 degrees.

## Why

I'd like to use images straight from my devices, without need to manually scrub EXIF data and then re-rotate each image.

## How

By adding Hugo's [auto orient](https://gohugo.io/functions/images/autoorient/) filter, wherever image is resized in layouts.

### DIFF
--- a/layouts/partials/article-link/card-related.html
+++ b/layouts/partials/article-link/card-related.html
@@ -24,7 +24,11 @@
         <div class="w-full thumbnail_card_related nozoom" style="background-image:url({{ .RelPermalink }});"></div>
         {{ end }}
       {{ else }}
-        {{ with .Resize "600x" }}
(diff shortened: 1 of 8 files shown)


nunocoracao:
Sorry @karosas not approving this one. Please submit a ticket to discuss implementation. This introduces a breaking change in behaviour without providing a config to control it. Also, without a ticket I don't know how valuable this change is - aka is this something just for you or for most users? 

Again thanks I really appreciate the contribution but can't merge this without understanding more about motivation and implementation.

PR #1461, opened 2024-05-11
✨ codeimporter shortcode

mnjm:
Added a shortcode (modified from mdimporter) for importing code files from external sources.


ddoddii:
Great feature ! Saves a lot of copy and pasting effort 👍

PR #1477, opened 2024-05-18
+sociallink shortcode

mnjm:
Added a `sociallink` shortcode that creates an anchor link for author socials from config file.

> New to Hugo and not entirely comfortable with it. Reviews appreciated. :)

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -339,7 +339,7 @@ In order to add images to the gallery, use `img` tags for each image and add `cl
 
 ## GitLab Card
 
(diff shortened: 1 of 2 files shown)


nunocoracao:
@mnjm closing this one. Please open an issue so that this gets discussed. Don't feel like adding more code to the theme without a clear use-case. This will increase the complexity of maintaining the project and it seems to be quite redundant with the already existing author blocks.

mnjm:
I also had second thoughts about this. I only use it to add socials on the lander page (like about) and sorta realized it might be redundant. Anyways

Thanks for the theme; Really appreciate your work . @nunocoracao

PR #1504, opened 2024-05-31
✨ Use markdown emojify instead of dump emojify

### LINKED ISSUE #1497: ✨ Emoji are rendered inside code blogs
**Describe the bug**

Using code blog inside markdown content files do not escape emoji conversion.

For example: ffmpeg -map 0:a:0 converts :a: to 🅰️

This is because hugo's emojify transform does not know about markdown code blogs, it was fixed on hugo's side by disabling their internal emojify transform and using the markdown's renderer one (see https://github.com/gohugoio/hugo/issues/4978 for more info).

Please note that disabling hugo's emojify is not possible (to my knowledge) since blowfish hardcodes `| emojify` calls everywhere. This means the config option `enableEmoji = false` does nothing.

**To Reproduce**
1. Create a blowfish blog
2. Add a markdown file with code blocks containing `:a:`
3. Open the rendered markdown file and observe the :a: instead of `:a:` inside the code block

**Expected behavior**
Not emojify code blocks.

**Screenshots**
image

For the following codeblock:

```bash
ffmpeg -i input.mkv -map 0:V:0 -c:v libx264 \
  -filter:v:0 scale=480:360  -b:v:0  800000 \
  -filter:v:1 scale=640:480  -b:v:1 1200000 \
  -filter:v:2 scale=1280:720 -b:v:1 2400000 \
  -var_stream_map "v:0,name:360p v:1,name:480p v:2,name:720p" \
  -f hls output.m3u8
```

**Hugo & Blowfish versions**
`hugo v0.126.1+extended linux/amd64 BuildDate=unknown VendorInfo=nixpkgs`
`github.com/nunocoracao/blowfish/v2 v2.68.0`

zoriya:
Fixes #1497 

Breaking: users need to enable native hugo emojies for them to show up.

This can be done by adding: `enableEmoji = true` inside the `config.toml`/`hugo.toml`

### DIFF
--- a/exampleSite/layouts/partials/recent-articles-demo.html
+++ b/exampleSite/layouts/partials/recent-articles-demo.html
@@ -37,7 +37,7 @@
     <a href="{{ $showMoreLinkDest }}">
         <button
             class="bg-transparent hover:text-primary-500 prose dark:prose-invert font-semibold hover:text-white py-2 px-4 border border-primary-500 hover:border-transparent rounded">
-            {{ i18n "recent.show_more" | markdownify | emojify }}
+            {{ i18n "recent.show_more" | markdownify }}
             </button>
     </a>
 </div>
--- a/layouts/_default/index.json
+++ b/layouts/_default/index.json
@@ -9,8 +9,8 @@
     "date" (.Date | time.Format (.Site.Language.Params.dateFormat | default ":date_long")) 
     "title" (.Title | emojify | safeJS)
     "section" ($section.Title | emojify | safeJS)
-    "summary" (.Summary | emojify | safeJS)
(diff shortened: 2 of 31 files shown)


MaikelChan:
I just noticed that in footer.html, line 30, `markdownify` was removed instead of `emojify`. Is it intended?

https://github.com/zoriya/blowfish/blob/9dc79407188adef714954ce463628f58e4725118/layouts/partials/footer.html#L30C14-L30C20

zoriya:
My bad, this is indeed a mistake.

nunocoracao:
@MaikelChan @zoriya Fixed in https://github.com/nunocoracao/blowfish/pull/1516

PR #1513, opened 2024-06-03
🎨 Improvements to image handling and link handling

wermos:
# Links

Due to the way the old shortcode was written, weird cosmetic errors like the one below were cropping up:
image

The input Markdown was:
```md
2. The ANTLR generated files generate huge diffs when changes are made to the grammar files. ([Here is an example.](https://github.com/sympy/sympy/pull/23535/files))
```

I changed the shortcode to not add that spurious space.
 
# Images

When I was investigating the HTML output for the image shortcodes, I noticed two things:
1. Usage of the image shortcodes introduced a lot of newlines into the output HTML. While it's not a big deal in production, it made debugging a little annoying. I removed all the whitespace from the image shortcodes.
2. Including a local SVG in the image shortcodes, when `disableImageOptimization` is `false`, would cause the Hugo build to fail because you cannot resize a SVG. I included a check for this edge case in the image shortcodes.

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -18,7 +18,7 @@ In addition to all the [default Hugo shortcodes](https://gohugo.io/content-manag
 <!-- prettier-ignore-start -->
 | Parameter   | Description                                                                                                                                                                                  |
 | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


michaeldorner:
Thanks for your efforts in this pull request, @wermos. I was recently playing around with SVG in Blowfish, but now I can't wait to see SVG support for the next release!

wermos:
Thank you for the kind words! I'm glad you found it useful 😄

Issue #1549, opened 2024-06-18
Codeimporter Shortcut doesn't handle URL encoding


ennnbeee:
**Describe the bug**
When using GitHub raw urls, with files containing spaces, the url contains '%20', when using this URL with the codeimporter shortcode, the page failes to render:

```
 render: failed to render pages: render of "page" failed: "C:\Source\github\blog\themes\blowfish\layouts\_default\single.html:83:13": execute of template failed at <.Content>: error calling Content: "C:\Source\github\blog\content\posts\Windows-CIS-Patching-Gaps-Part3\index.md:212:1": failed to render shortcode "codeimporter": failed to process shortcode: "C:\Source\github\blog\themes\blowfish\layouts\shortcodes\codeimporter.html:4:47": error calling resources.GetRemote: failed to parse URL for resource https://raw.githubusercontent.com/ennnbeee/mve-scripts/main/Intune/Remediation/CIS/L1/CIS%!!(MISSING)((MISSING)L1)%!!(MISSING)S(MISSING)ystem%!!(MISSING)S(MISSING)ervices%!!(MISSING)-(MISSING)%!!(MISSING)W(MISSING)indows%!I(MISSING)ntune%!!(MISSING).(MISSING)1_Detection.ps1: parse "https://raw.githubusercontent.com/ennnbeee/mve-scripts/main/Intune/Remediation/CIS/L1/CIS%!!(MISSING)((MISSING)L1)%!!(MISSING)S(MISSING)ystem%!!(MISSING)S(MISSING)ervices%!!(MISSING)-(MISSING)%!!(MISSING)W(MISSING)indows%!I(MISSING)ntune%!!(MISSING).(MISSING)1_Detection.ps1": invalid URL escape "%!!(MISSING)(" 
 ```

**To Reproduce**
```
{{< codeimporter url="https://raw.githubusercontent.com/ennnbeee/mve-scripts/main/Intune/Remediation/CIS/L1/CIS%20(L1)%20System%20Services%20-%20Windows%2011%20Intune%203.0.1_Detection.ps1" type="PowerShell" >}}
```

**Expected behavior**
Expect the shortcode to handle the url provided, without having to replace the encoded characters.

mnjm:
Opened a PR with fix; [here](https://github.com/nunocoracao/blowfish/pull/1550)

nunocoracao:
Thanks @mnjm and thanks @ennnbeee for bringing this up

PR #1550, opened 2024-06-19
🐛 fix: https://github.com/nunocoracao/blowfish/issues/1549

mnjm:
Fix https://github.com/nunocoracao/blowfish/issues/1549


nunocoracao:
Thanks @mnjm

PR #1577, opened 2024-06-26
:sparkles: Add data definition for sharing on Bluesky

mpoore:
Added:

- Bluesky definition in data/sharing.json
- Bluesky to exampleSite params.toml
- Alt text to i18n/en.yaml

Enables the addition of "bluesky" to the default article sharing links using the Action Intent link defined here: https://docs.bsky.app/docs/advanced-guides/intent-links

### DIFF
--- a/data/sharing.json
+++ b/data/sharing.json
@@ -38,5 +38,10 @@
     "icon": "telegram",
     "title": "sharing.telegram",
     "url": "https://t.me/share/url?url=%s&resubmit=true&title=%s"
+  },
+  "bluesky": {
+    "icon": "bluesky",
+    "title": "sharing.bluesky",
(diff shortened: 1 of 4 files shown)


nunocoracao:
@mpoore thanks for this

Two things that I'll fix after the merge
- the update to the config files is missing - aka the files that new users start with
- I only update the version number upon release - aka dev merge to main which is not the case

mpoore:
> @mpoore thanks for this
> 
> 
> 
> Two things that I'll fix after the merge
> 
> - the update to the config files is missing - aka the files that new users start with
> 
> - I only update the version number upon release - aka dev merge to main which is not the case

Thank you @nunocoracao 
I didn't include the default config only because I couldn't complete the i18n for all languages. 😁

Issue #980, opened 2023-10-01
⚙️ minimize repo size


fuse314:
**Describe the bug**
My CI script fetches the submodule (blowfish theme) every time the build process is run.
The repo size is over 700MB, mostly due to the folders `.git`, `exampleSite` and `public`

**To Reproduce**
1. Check out a fresh copy of the repository `time git clone https://github.com/nunocoracao/blowfish blowfish-test`
  On my reasonably fast CI-VM, the process takes around 21 Seconds.
3. Check the directory sizes `du -h --max-depth 1 ./blowfish-test/`
   ```
   32K     ./blowfish-test/config
   60K     ./blowfish-test/.github
   12K     ./blowfish-test/archetypes
   124K    ./blowfish-test/static
   221M    ./blowfish-test/exampleSite
   24K     ./blowfish-test/data
   326M    ./blowfish-test/.git
   4.7M    ./blowfish-test/assets
   169M    ./blowfish-test/public
   100K    ./blowfish-test/i18n
   16M     ./blowfish-test/images
   476K    ./blowfish-test/layouts
   736M    ./blowfish-test/
   ```
   Total size: 736MB, biggest folders: `.git` (326MB), `exampleSite` (221MB) and `public` (169MB).
3. Check the directory sizes of exampleSite `du -h --max-depth 1 ./blowfish-test/exampleSite/`
   ```
   121M    ./blowfish-test/exampleSite/content
   28K     ./blowfish-test/exampleSite/config
   8.0K    ./blowfish-test/exampleSite/archetypes
   84M     ./blowfish-test/exampleSite/resources
   16K     ./blowfish-test/exampleSite/data
   17M     ./blowfish-test/exampleSite/assets
   24K     ./blowfish-test/exampleSite/layouts
   221M    ./blowfish-test/exampleSite/
   ```
   Biggest folders: `content` (121M), `resources` (84MB) and `assets` (17MB).

**Expected behavior**
The build performance for CI builds -- and the general size of the repository -- could be improved by one of the following points:
1. Remove the `public` folder from repository, it contains the generated webpage from `exampleSite`, that could easily be regenerated by running `hugo`. (removes 169MB - also remove the git history of this folder to shrink the `.git` folder.)
2. Put the `exampleSite` folder in its own submodule, so a shallow checkout of the theme would be around half of the current size (removes 221MB, rewrite the git history to shrink the `.git` folder as well.)
[...]

PR #1588, opened 2024-07-02
Remove unneeded files in lib/mermaid

wtchangdm:
Some context: https://github.com/nunocoracao/blowfish/issues/980#issuecomment-2198018085

There was a finding that `assets/lib/mermaid` takes ~90MiB size. After digging the [history](https://github.com/nunocoracao/blowfish/commits/main/assets/lib/mermaid) a little, I am still curious why all files were included again:

1. It was only `mermaid.min.js` until https://github.com/nunocoracao/blowfish/commit/7740790c282781533747b92b7e794245b343e02f.
2. Mermaid was removed from dependencies at https://github.com/nunocoracao/blowfish/commit/7e41c9cac0e9249dfde6d85990878ee7a36b8418
3. Mermaid was re-introduced in https://github.com/nunocoracao/blowfish/commit/5a11829222aa535ec007ddfaf82c632b558937d2, with full folder copy command in `package.json` along with many other files

However, after checking the code that includes mermaid:

https://github.com/nunocoracao/blowfish/blob/c6c2689195eb47241c613a1e4b8d0828ebf3c6e4/layouts/partials/vendor.html#L6-L10

Looks like what we need here is only `mermaid.min.js`.

Therefore I created this PR to remove the unneeded files and reduce the folder size from 90MiB to 3.3MiB.

This should benefit users using blowfish as hugo module and the ones who only shallow-clones the repo.

@nunocoracao please let me know if I missed anything.

Thanks!


wtchangdm:
Oops, just found out I should fork from `dev` branch and send PR to it instead.

wtchangdm:
The last two commits were caused by mistake as I was testing the result hugo module but changed the module name on wrong branch and have fixed it.

PR #1593, opened 2024-07-05
✨ umami analytics support data-domains

Weaxs:
umami analytics support [data-domains](https://umami.is/docs/tracker-configuration#data-domains). 

this param can make umami script only run specific domains.


nunocoracao:
Thanks @Weaxs

PR #1597, opened 2024-07-08
add `bilibili` shortcode and `ruby` extended syntax

imagicw:
1. use `{{<bilibili id>}}` to add bilibili video

2. support `ruby` syntax:

enable in `Params` configs: `ruby = true`

example:

`[関連]^(かんれん)`  

image

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -14,6 +14,25 @@ body.zen-mode-enable {
   }
 }
 
+.bilibili {
+  position: relative;
+}
+
+.bilibili::after {
+  content: "";
+  display: block;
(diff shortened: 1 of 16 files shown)


nunocoracao:
Sorry @imagicw I am closing this PR. The following reasons
1. it doesn't seem to be related to tasks in the backlog - adding these things were never discussed
2. the PR adds multiples features 
3. too long for review
4. it seems to lack documentation

Please open issues so that the community has a way to discuss these topics and let's have one feature being added (with docs) per PR.

lightmon233:
I think that ruby annotation support is worth consideration.

Issue #1639, opened 2024-07-25
✨ Support Seline Analysis


Weaxs:
**Is your feature request related to a problem? Please describe.**
Support https://seline.so/

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1640, opened 2024-07-26
✨ Support Seline Analysis 

Weaxs:
https://github.com/nunocoracao/blowfish/issues/1639

- Support [Seline Analysis](https://seline.so/)
- Umami Analysis add `enableTrackEvent`


exampleSite/content/docs/configuration/index.it.md
```diff
@@ -323,6 +323,14 @@ Many of the article defaults here can be overridden on a per article basis by sp
...
+| `umamiAnalytics.enableTrackEvent`    | true      | When set to `true` track event will add automatically. If you do not want to add track event, set it to `false`.                                                            |
+
+### Seline Analytics
+
+| Name                              | Default   | Description                                                                                                                               |
+|-----------------------------------|-----------|-------------------------------------------------------------------------------------------------------------------------------------------|
+| `selineAnalytics.token`           | _Not set_ | The token generated by Seline Analytics for the website. Refer to the [Analytics docs]({{< ref "partials#analytics" >}}) for more details. |
+| `selineAnalytics.enableTrackEvent` | true      | When set to `true` track event will add automatically. If you do not want to add track event, set it to `false`.                          |
```
ragibson, line 333: These should definitely default to false or not set to match all the other analytics engines. Otherwise, you're enabling it for every user by default.

exampleSite/content/docs/configuration/index.md
```diff
@@ -318,12 +318,19 @@ Many of the article defaults here can be overridden on a per article basis by sp
...
+| `umamiAnalytics.enableTrackEvent`    | true      | When set to `true` track event will add automatically. If you do not want to add track event, set it to `false`.                                                            |
 
+### Seline Analytics
+
+| Name                              | Default   | Description                                                                                                                               |
+|-----------------------------------|-----------|-------------------------------------------------------------------------------------------------------------------------------------------|
+| `selineAnalytics.token`           | _Not set_ | The token generated by Seline Analytics for the website. Refer to the [Analytics docs]({{< ref "partials#analytics" >}}) for more details. |
+| `selineAnalytics.enableTrackEvent` | true      | When set to `true` track event will add automatically. If you do not want to add track event, set it to `false`.                          |
```
ragibson, line 333: These should definitely default to false or not set to match all the other analytics engines. Otherwise, you're enabling it for every user by default.

layouts/partials/analytics/umami.html
```diff
@@ -10,6 +10,7 @@
 </script>
 {{ end }}
 
+{{ if .Site.Params.umamiAnalytics.enableTrackEvent | default true }}
```
ragibson, line 13: Same comment here -- I don't think these should be defaulting to enabled.

ragibson (changes requested):

exampleSite/content/docs/configuration/index.md
Weaxs: > These should definitely default to false or not set to match all the other analytics engines. Otherwise, you're enabling it for every user by default.

In my opinion `enableTrackEvent=true` is recommended for optimal data tracking and display.

If you want to customize event tracking, might set it to `false`, and then add manually invoking `umami.track()` or `seline.track()` for specific events. 
However, for most users this customization is unnecessary.

ragibson: I see, there's an outer guard here in `analytics/main.html`.

PR #1646, opened 2024-07-29
✨ add SVG support for logos

ldericher:
This is my proposal to use an SVG as `.Site.Params.Logo`. 

Additions, hints and corrections are welcome!

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -3163,6 +3163,13 @@ body button {
   width:1em
 }
 
+/* Scale SVG logos to appropriate size */
+
+.logo svg {
+  height:5rem;
+  width:5rem
+}
+
 /* Search */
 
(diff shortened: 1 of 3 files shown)


ldericher:
I can see my PR will not handle `.Site.Params.SecondaryLogo` correctly as it is replaced using JS in https://github.com/nunocoracao/blowfish/blob/18ed67b2e613e2803d05b857327d73e408d90d53/assets/js/appearance.js#L80

This might be irrelevant aka. "you should create a reactive SVG using CSS in that case"

ldericher:
Appreciate the merge, but this was WIP and I'm gonna propose another change instead. Rationale:

1. This does not handle the "secondary" logo at all ie. breaks the parameter.
2. The raw SVG logo data will be injected into every HTML file, therefore will not be cached!

PR #1659, opened 2024-07-31
✨ Add 'Reply by Email' feature

### LINKED ISSUE #1611: ✨ Add 'Reply by Email' link after blog post
Not really. This is meant to be an alternative to having a comment section or utilising platforms like Disqus. 

I've created a solution for this where one adds the email field under author details in the `language.en.toml` file and this will be used when creating the link throughout the site. The link itself will contain an email subject `Reply to [Post Title]` for the author to know which post the reader is replying to at a glance when it lands in their inbox. 

Manually adding the link after each post or utilising archetypes but both feel less clean to me especially when someone changes their preferred email.

**Additional context**
I had initially read this post on [niqwithq.com](https://niqwithq.com) called [Reply by email](https://niqwithq.com/posts/reply-by-email) but didn't give it much thought until quite recently. It feels to me like a simple solution of getting feedback without having a comments system. Additionally, I'm also seeing it in different forms on more and more personal blogs, perhaps others might find it useful as well. What do you think about this @nunocoracao ?

As mentioned, I have a branch for my site where I've created a functional version of it (screenshot below). I wanted to now make it a param but I've never done that before so I wanted to get initial feedback on the feature's necessity before I put in extra work on it/create a PR. 

image

**Update**: made it into a param locally on my side that one can override in front matter

insidemordecai:
Closes #1611 where I suggest including a low maintenance `Reply by Email` feature.

I have set it to `false` by default to match the current version of Blowfish and it's flexible through configuration via the front matter. Additionally, I have updated the documentation to reference the feature.

Additions, corrections, and any other suggestions are welcome.

For reference, here's how it looks on a live dev deployment of my site: [https://967f13f0.insidemordecai.pages.dev/](https://967f13f0.insidemordecai.pages.dev/)


nunocoracao:
I think the partial needs a default in itself it there is no value in the params... but I'll fix it

insidemordecai:
thanks @nunocoracao I appreciate the merge

PR #1679, opened 2024-08-10
👾 Correct time format

sculptor189:
Modified incorrect time format.


ldericher:
I'm curious: What does this do/fix?

sculptor189:
With the original code, the format was incorrect and had errors when validated on W3C validator. So, I specified the format.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

PR #1680, opened 2024-08-11
Revert "✨ add SVG support for logos"

ldericher:
Reverts nunocoracao/blowfish#1646

> Rationale:
>     1. This does not handle the "secondary" logo at all ie. breaks the parameter.
>     2. The raw SVG logo data will be injected into every HTML file, therefore will not be cached!

:warning: **Please do not merge yet, this is WIP** :warning:

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -3163,13 +3163,6 @@ body button {
   width:1em
 }
 
-/* Scale SVG logos to appropriate size */
-
-.logo svg {
-  height:5rem;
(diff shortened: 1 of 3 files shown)


ldericher:
Question: What do the `width` and `height` params on the original `img.logo` DOM element even do that isn't already done in CSS?

nunocoracao:
@ldericher I am closing this PR as it is not ready to merge. Please work off your branch and open the PR once it's ready to merge. If you need help let's use the respective issue

PR #1787, opened 2024-10-10
fix: exclude pages with "xml" = false or "externalUrl".

fuse314:
Pages that include the `"xml" = false` attribute are excluded from the XML-Sitemap, but they still appeared in the RSS-feed.
Pages that include a `externalUrl` that starts with `http://` or `https://` are excluded from the XML-Sitemap, but they still appeared in the RSS-feed.

This PR implements the same filters used in sitemap.xml in the RSS-feed template.


nunocoracao:
Sorry @fuse314 closing this one. In my case, for example, I do want external links to popup in RSS feed and exclude them from the sitemap.

PR #1803, opened 2024-10-18
Update users.json

mer0x:
add merox.dev website


nunocoracao:
Closing because the PR messes up other users... I will add manually

PR #1822, opened 2024-10-28
feat: option to plainify summary

vehagn:
In the release of [Hugo v0.134](https://github.com/gohugoio/hugo/releases/tag/v0.134.0) they changed the way summaries are done. Not plainifying the summaries leads to layout issues in lists, this fixes that


nunocoracao:
Thanks @vehagn I was looking for this solution. I am going to use it, except without a params config. This should be the default behaviour.

designproworks:
Awesome thank you everyone. I will be updating soon with meta data and google anlytics. https://designproworks.com

PR #1986, opened 2025-02-17
Linktree page

fguisso:
Usage:
```
---
layout: linktree
links:
  - "Extra links": "https://example.com"

---
```

This layout show all your social links from your `.Params.Author.links` and show extra links from the actual page.

There is my [Live Demo](https://guisso.dev/links), but it's not equal to this PR, I have customized with more [code](https://github.com/fguisso/fguisso.github.io/blob/486167e62d4d9dff8daee4d8d04021eedd026402/layouts/partials/linktree_list.html#L19).

Here is a live demo using de [blowfish exampleSite](https://guisso.dev/blowfish/links/)

image

### DIFF
--- a/exampleSite/content/docs/content-examples/index.md
+++ b/exampleSite/content/docs/content-examples/index.md
@@ -90,6 +90,64 @@ _In this example, the special `cascade` parameter is being used to hide the read
 
 The [samples section]({{< ref "samples" >}}) of this site is an example of a list page.
 
+### Linktree Page
+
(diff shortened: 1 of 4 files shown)


nunocoracao:
Thanks for this @fguisso I'll be rejecting this PR though.
Something like this would need an issue an some thoughts. Don't intend to add more code to maintain at this stage unless some users want it. Also, the design doesn't really align with the rest of the theme.

Issue #1663, opened 2024-08-02
🐛 Footer line issue & Author info disappears in ZenMode


DarkXero-dev:
Hello

Using latest Blowfish version I noticed when browser is in full screen the separator line in the footer does not stretch all the way.. Look at image below...

![line](https://i.imgur.com/DA4zZg4.png)

Minor issue of course but kind of annoying. Also when ZenMode is active Author info disappears.

hayduke-io:
Just posted a PR for a fix to the divider line in zen mode issue. 

The article footer was left out of the class transforms of the zen-mode js file. Small additions to /layouts/_default/single.html and /assets/js/zen-mode.js 

The author issue is separate and I'm still looking into it. Recommend it as its own issue. The author info in zen mode behavior is working as intended. Per /assets/css/components/zen-mode.css,
`body.zen-mode-enable {
    #bmc-wbtn, .author {
        display:none !important;
    }
}`

However, when edited or even outright removed from codebase, the behavior persists. 

Also likely more cleanup on the single.html is needed. The single.footer border divider is tied to Params.showComments value. But Blowfish doesn't use that parameter. 

Let me know if I have any of that wrong. I'm new to contributing and have been learning Hugo and Blowfish by just sitting in the code and docs.

cheers

PR #2001, opened 2025-02-25
:bug:Fix zen mode footer line bug

hayduke-io:
Re Issue #1663

Fix for article footer border line not expanding when in zen mode


nunocoracao:
@coderabbitai review

PR #2010, opened 2025-02-27
Remove unused option in `timeline` shortcode

wermos:
The `text` option was undocumented and also not used anywhere in the shortcode.


nunocoracao:
@coderabbitai review

PR #2013, opened 2025-03-08
✨ Create gist.html

michaeldorner:
Since the gist shortcode was deprecated in v0.143.0 and will be removed in a future release of Hugo, I followed the instructions provided by [Hugo](https://gohugo.io/shortcodes/gist/).


nunocoracao:
@michaeldorner thanks for the PR. Could you add the documentation to work with this?

michaeldorner:
Sure! But since I just followed Hugo’s instructions to avoid the deprecation warning, I’m not sure what specifically needs to be documented. Is there something specific you’d like covered?

nunocoracao:
Since the gist shortcode stops being a Hugo primitive and becomes a theme one we should document how it can be used. Adding something to the shortcodes page in docs explain what it is, table of parameters, and an example would help users get started.

PR #2014, opened 2025-03-08
Update to newest Hugo version

michaeldorner:
To avoid the Hugo warning `Module "blowfish" is not compatible with this Hugo version: 0.87.0/0.143.1;`, I updated the `max` version to `0.145.0`.


nunocoracao:
Dup PR

PR #2049, opened 2025-04-09
Update users.json

papo1011:
add new user blog.francescopapini.com


nunocoracao:
closing PR - should be added to end of the list
Will add manually

PR #2104, opened 2025-05-03
🐛 Fix: handle uppercase values in templates.Exists lookup

ZhenShuo2021:
Due to a bug in Hugo v0.146.0, templates.Exists fails to resolve templates with uppercase characters. This patch adds fallback checks using lowercase names to ensure proper resolution (e.g. thumbAndBackground).


ZhenShuo2021:
Fixed in hugo [v0.147.2](https://github.com/gohugoio/hugo/releases/tag/v0.147.2)

PR #2116, opened 2025-05-06
✨ Feat: support extend-head access .Page (#1781)

### LINKED ISSUE #1781: Ability to access article frontmatter / `Page` object in `extend-head.html`

As far as I'm aware, it's currently not possible to access the current page's `Page` object (and by extension, any page parameters from frontmatter) in `extend-head.html`. This makes it impossible to perform any page-level logic.

For example, in my use case, I wish to add something in my `extend_head.html` which relies on a condition such as `if page.HasShortcode "foo_bar"`, but (again, as far as I'm aware?) this is not possible with the way `extend-head.html` is currently implemented, since Blowfish [only passes `.Site` into the partial](https://github.com/nunocoracao/blowfish/blob/main/layouts/partials/head.html#L124).

It would open up a lot of possibilities and be super handy to be able to access the current page from `extend-head.html`.

Given what I wish to do, I don't believe there is any alternative, but I'm open to correction!

**Additional context**

This is the same issue as #1425, which appears to be have been prematurely closed due to a misunderstanding.

ZhenShuo2021:
Fix #1781: Use extend-head-uncached.html for backward compatibility.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -128,6 +128,10 @@
   {{ if templates.Exists "partials/extend-head.html" }}
   {{ partialCached "extend-head.html" .Site }}
   {{ end }}
+  {{/* Uncached extend head - e.g. {{ with .Page.HasShortcode "gallery" }} do something {{ end }}  */}}
+  {{ if templates.Exists "partials/extend-head-uncached.html" }}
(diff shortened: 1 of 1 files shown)


nunocoracao:
Thanks @ZhenShuo2021 all this work has been amazing. 
Could you add docs for this one?

PR #2133, opened 2025-05-11
Feat: Adds ability show only the category instead of all tags

timcappalli:
`showTaxonomies` shows both tags and categories. Tags can get a bit busy depending on the count.

This change adds the ability to show only the category.


nunocoracao:
Thanks @timcappalli if you could open a PR to update the relevant docs for this

PR #2172, opened 2025-05-27
🔧 Chore: adjust develope command

ZhenShuo2021:
## This PR affects codes uses `npm run example`!

I realized that the `example` command was not clear enough. After just two weeks, I had to spend time figuring out what I meant.

Now, both the production and environment commands are clearly defined, and the language disabling feature can be configured via an environment variable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@
 
 - Great! Open a new GitHub pull request with the patch.
 - All development occurs on the `dev` branch and new PRs should be forked from here.
-- The command `npm run example` can be used to test local changes using the example site.
+- The command `npm run example:production` can be used to test local changes using the example site.
 - Ensure the PR description clearly describes the problem and solution. Include the relevant issue number, if applicable.
 - **Before submitting**, check the [coding conventions](#coding-conventions) section below to learn more about coding and commit message expectations.
 
--- a/exampleSite/config/_default/hugo.toml
+++ b/exampleSite/config/_default/hugo.toml
@@ -70,11 +70,19 @@ googleAnalytics = "G-PEDMYR1V0K"
       type = 'fragments'
       weight = 10
 
-# Render only specific parts of the site for faster development. Use: hugo --renderSegments site_core
+# Render only specific parts of the site for faster development. 
+# Note that there are more files are ignored in development/hugo.toml
+#
+# Usage: `hugo --renderSegments core` or `hugo --renderSegments core,en_only`
 [segments]
-  [segments.site_core]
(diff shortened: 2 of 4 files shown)


nunocoracao:
Why remove the example entry?
I would prefer

- example:dev --> example (as of today i.e. dev is the default reason to run this command so no need to waste more chars on that)
- example:dev:core ---> example:core
- example:production ----> love it as-is

ZhenShuo2021:
Just to provide a bit of context first:

> In the dev environment, folders such as `users` and `guides` are not rendered to speed up the local development server. This can result in confusing behavior, like clicking on links that appear to do nothing, which might be misleading during development or review.

So using `example:dev` is meant as a soft hint that the environment is in dev mode. Since the actual environment config is nested under `exampleSite/config/config/development`, it's not obvious at a glance.

This also gives a clearer mental model. If it's not `:production`, then it's understood to be dev. With just `example`, it could be less clear what mode is actually being used.

That's the thinking behind the current naming.

ZhenShuo2021:
This was the naming pattern I originally considered:

```
"dev:css": "NODE_ENV=development tailwindcss ...",
"dev:css:windows": "set NODE_ENV=development tailwindcss ...",
"build:css": "NODE_ENV=production tailwindcss ...",
"build:css:windows": "set NODE_ENV=production tailwindcss ...",
"build:hugo": "hugo --minify ...",
"dev:hugo": "hugo server ...",
"dev:hugo:core": "hugo server --renderSegments core ...",
"dev:hugo:production": "hugo server -e production ...",
```

But changing everything at once might introduce more issues. Since I already made some changes and immediately regretted them, I chose not to include this version in the current PR. It can be revisited separately if needed.

nunocoracao:
Fair. I still prefer the

> Why remove the example entry? I would prefer
> 
> * example:dev --> example (as of today i.e. dev is the default reason to run this command so no need to waste more chars on that)
> * example:dev:core ---> example:core
> * example:production ----> love it as-is

There's no good reason to change this apart from the theorectical thought you've put into it - which I totally understand and would be ok with if this was a new project. Since it's not and changing some of those have impacts that are not easily estimated while they don't bring much value. I am ok in creating new ones following that logic - I am not ok with changing the existing ones - specifically example.

ZhenShuo2021:
I agree, but using `example` will behave differently than before. Originally, `example` meant production; this PR changes it to development, and the original becomes `example:production`.

How would you like to proceed? I can create a new commit.

nunocoracao:
I think it's ok that example becomes the dev command and production requires an extra step
If you can open a new commit good - otherwise I can fix this after in the dev branch - up to you

PR #2174, opened 2025-05-27
📝 Docs: rename partials to extensions

ZhenShuo2021:
(Base on #2173)

The current documentation on "Partials" can be confusing for first-time Hugo users, myself included. The content actually focus on extend site functionality, such as integrating analytics, comments, icons, and customizing the head or footer. The only relationship to the `partial` is these features uses `partial` function in code.

Thus, rename it to "Extensions" might be more appropriate.


nunocoracao:
This uses the same naming convention as Hugo - might seem harder at the beginning but it's aligned with Hugo docs which makes things easier in the future. Closing PR

PR #2176, opened 2025-05-27
🎨 Style: add formatter config

ZhenShuo2021:
This PR introduces a formatter configuration aimed at significantly improving the development experience.

## .editorconfig

This file enables IDEs to apply the correct tab indentation and display nested statements properly. The configuration is based on the setup described in `CONTRIBUTING.md`.

## Contributing.md Updates

I noticed that around 99% of contributors do not use gitmoji in commit messages. Therefore, I've changed the it to PR instead.

> Check the usage with `git log --perl-regexp --author='^(?!.*(nunocoracao@gmail.com|dependabot\[bot\]|ZhenShuo)).*' --pretty=format:'%C(cyan)%an%C(reset): %s %C(dim)%ad%C(reset)' --date=format:'%Y-%m'`

## Formatter Configuration

This is the core of the PR. Without any formatter settings, development becomes a harrowing.

While `CONTRIBUTING.md` states that we should use 2-space indentation, but actually we have the follow formats:

1. [1-space indent](https://github.com/nunocoracao/blowfish/blob/0b06a64139beba6287e7685f4c810ad4ff772fde/layouts/partials/hero/background.html)
2. 2-space indent in most files
3. At least we don't have 3-space indent (that would be pure evil)
4. [4-space indent](https://github.com/nunocoracao/blowfish/blob/0b06a64139beba6287e7685f4c810ad4ff772fde/layouts/partials/home/background.html)
5. [No indentation](https://github.com/nunocoracao/blowfish/blob/0b06a64139beba6287e7685f4c810ad4ff772fde/layouts/partials/head.html), which makes it hard to identify closing statements, especially with mixed usage of `with`, `range`, and `if`, and inconsistent use of one-liner `if` statements. The worst part is these problems happens together.
6. [Mixed indentation](https://github.com/nunocoracao/blowfish/blob/0b06a64139beba6287e7685f4c810ad4ff772fde/layouts/partials/article-link/simple.html)
7. [Misleading indentation due to typos](https://github.com/nunocoracao/blowfish/blob/0b06a64139beba6287e7685f4c810ad4ff772fde/layouts/partials/header/basic.html#L144-L145)

With a proper Prettier configuration, all of these pain can be cured. The configuration is largely based on [Hugo's official repo](https://github.com/gohugoio/hugo/blob/master/docs/.prettierrc), and files with invalid nodes are excluded. This allows us to run the following command to automatically fix indentation issues:

```sh
npx prettier --write --list-different .
```

### Why This Setup

The current Prettier package is essentially unused, as it has no configuration. For HTML files, it cannot recognize Go template indentation. For Markdown files, it also breaks shortcode usage. Adding `<!-- prettier-ignore-start -->` before every shortcode is not a practical solution.

### Limitations

The format config is not perfect. Firstly, it cannot format markdown files (I suggest to use markdownlint instead). Second, we typically prefer to write HTML tags like this:

```html
<!-- we want "</script> takes an independent new line -->  

<script
  type="text/javascript"
  src="{{ $jsZenMode.RelPermalink }}"
  integrity="{{ $jsZenMode.Data.Integrity }}">
</script>
```

However, the prettier plugin will format it to

```html
<script
  type="text/javascript"
  src="{{ $jsZenMode.RelPermalink }}"
  integrity="{{ $jsZenMode.Data.Integrity }}"></script>
```

Additionally, the [prettier-plugin-go-template](https://github.com/NiklasPor/prettier-plugin-go-template#readme) hasn't been updated for a while, so we may encounter minor bugs that could take time to identify and resolve. After some research, this plugin is the only plugin able to format HTML with Hugo.

### Notes

With file formatting, no PR will be able to merge. I suggest merging the formatted version into the main branch without creating a new release, so both users and developers can continue working smoothly.


nunocoracao:
>With file formatting, no PR will be able to merge. I suggest merging the formatted version into the main branch without creating a new release, so both users and developers can continue working smoothly.

Could you clarify this?

ZhenShuo2021:
The sentence "With file formatting, no PR will be able to merge." actually refers to PR #2177, which applies Prettier formatting. Its inclusion in PR #2176 was a mistake. I've add this sentence to the decription in #2177.

PR #2194, opened 2025-06-02
🔒 Add config parameter fingerprintAlgorithm

servedsmart:
This parameter allows users to choose their algorithm for html integrity checks. This mainly affects Hugo's fingerprint

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -29,6 +29,8 @@ disableTextInHeader = false
 # smartTOC = true
 # smartTOCHideUnfocusedChildren = true
 
+fingerprintAlgorithm = "sha512" # Valid values are "sha512" (default), "sha384", "sha256"
+
 giteaDefaultServer = "https://git.fsfe.org"
 forgejoDefaultServer = "https://v8.next.forgejo.org"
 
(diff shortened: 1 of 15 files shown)


servedsmart:
I myself need this feature because I am trying to generate a CSP that I send to Cloudflare API. Cloudflare apparently has some limit on header size or CSP length, therefore I need to save some bytes. This usecase might also apply to others.

servedsmart:
I am not sure if this will catch all fingerprints, but I didn't find any other mentions of sha512 in html files.

nunocoracao:
>I myself need this feature because I am trying to generate a CSP that I send to Cloudflare API. Cloudflare apparently has some limit on header size or CSP length, therefore I need to save some bytes. This usecase might also apply to others.

Have you tried it out locally and does it solve your use-case?

I am keen on approving this but I am also weary of adding even more configurations to the theme as it stands.

servedsmart:
> > I myself need this feature because I am trying to generate a CSP that I send to Cloudflare API. Cloudflare apparently has some limit on header size or CSP length, therefore I need to save some bytes. This usecase might also apply to others.
> 
> Have you tried it out locally and does it solve your use-case?
> 
> I am keen on approving this but I am also weary of adding even more configurations to the theme as it stands.

Yes, it reduces my CSP by around 20% since I was already using sha256 for inline scripts. I am also using this in html that I've added myself and it works flawlessly.

Thank you very much for your initial review. I understand your point about even more config options, but at least this shouldn't be too much of a burden to maintain. People would however always have to use `(.Site.Params.fingerprintAlgorithm | default "sha512")` when setting up integrity checks.

Issue #2195, opened 2025-06-02
Don't create inline style in javascript to allow hashes in CSP


servedsmart:
**Is your feature request related to a problem? Please describe.**

At the moment, as far as I can tell, it is impossible to use a somewhat strict CSP that relies on hashes. There are a lot of inline scripts and styles in html code, but those can be hashed somewhat easily with a script.
There however seems to be [at least one script](https://github.com/nunocoracao/blowfish/blob/main/assets/lib/zoom/zoom.min.js) that creates dynamic inline css styles. One would have to create a hash for every possible returned inline style by that script to be able to allow it in a CSP.

It is possible that my conclusion is due to an oversight on my part, but I am pretty sure that this basically doesn't allow users to use hashes in their CSP.

**Describe the solution you'd like**

Find another way to implement the mentioned script's functionality without creating dynamic inline styles in javascript.
[This](https://stackoverflow.com/questions/42401952/inline-style-error-with-content-security-policy-and-javascript) is a discussion on a similiar issue. I however don't know if the mentioned solution would be of any use here since it seems to do something slightly different.

**Describe alternatives you've considered**

Find a way to implement this without creating any inline styles.

**Additional context**

I might add console errors and my CSP here later, but I don't have a lot of time at the moment. Sorry for the very brief explainations.

servedsmart:
I can confirm that the pr has solved the CSP issues caused by `zoom.min.js`.

<img src="https://github.com/user-attachments/assets/16245d9f-bd9c-411a-be29-f68a1ebd0561" height="400" title="Screenshot Chrome Dev Tools" alt="Screenshot of Chrome Dev Tools showing 2 blocked styles"/>

These are the only remaining errors, at least on the root. Before that it reposted 2 errors originating from `zoom.min.js`, I sadly didn't get to screenshot those and it is somewhat of a hassle to recreate.

EDIT: It seems like this actually fixes all similar issues. What I am encountering as errors here is just a mistake in my processing of inline styles/scripts where not all actually get selected correctly.

PR #2196, opened 2025-06-03
:lipstick: Append stylesheet instead of inline style in medium-zoom

servedsmart:
This should fix the following issue: https://github.com/nunocoracao/blowfish/issues/2195
css sourced via: https://raw.githubusercontent.com/francoischalifour/medium-zoom/refs/heads/master/src/medium-zoom.css
js sourced and modified via: https://cdn.jsdelivr.net/npm/medium-zoom/dist/medium-zoom.js

I'm not certain how you handle external dependencies. I also modified the package files just to remove the now unneeded dependency and sourced it manually.
I have never used external libraries in javascript, but I don't think that this can easily be integrated with as far as I know npm (if you are using that :D)


servedsmart:
Please note that I have made use of the new config option from this pr: https://github.com/nunocoracao/blowfish/pull/2194

I can of course just use "sha512" if that will not be merged. I just didn't want to make changes later.

servedsmart:
In theory one could also merge this upstream but since this very much is backwards incompatible and makes it more tedious to integrate `medium-zoom.js` I don't think it is even worth asking them.

servedsmart:
Fixes: #2195 

Not sure why there is no relation created, but it should be obvious by the mentions.

servedsmart:
I can also move `medium-zoom.js` to `assets/js/` and move `medium-zoom.css` somewhere else if that is needed. Just didn't know where to put them.

nunocoracao:
I would prefer to manage this using npm - closing the PR

servedsmart:
> I would prefer to manage this using npm - closing the PR

@nunocoracao Okay, I understand this. The changes aren't too hard to just implement outside of the theme so that is completely fine with me!
What do you think of [this discussion](https://github.com/nunocoracao/blowfish/discussions/2198) I have started? I would be willing to work on this if people familiar with the project approve of these changes. It really shouldn't be too drastic since inline styles can probably be replaced by `class` and script tags seem easy to merge.

Issue #2093, opened 2025-05-02
Figure Shortcut broken when default=true


schneekluth:
**Describe the bug**
When using the in-built [Figure Shortcut](https://blowfish.page/docs/shortcodes/#figure) with setting `default=true` Hugo fails to build the site and complains about a missing partial:

`error calling partial: partial "shortcodes/figure.html" not found`

**Screenshots**

![Image](https://github.com/user-attachments/assets/63c0a47a-e667-40f4-a875-c084e2e1dc6c)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version:  24H2

**Hugo & Blowfish versions**
Which version of Hugo and Blowfish are you using?

- Hugo Version: 0.174.1
- Blowfish Version: 2.85.0

schneekluth:
I was able to fix this with the instructions from this issue https://github.com/nunocoracao/blowfish/issues/2052.

The missing file [`figure.html`](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_shortcodes/figure.html) needs to be copied to `layouts/shortcodes`. After that everything worked as expected.

PR #2055, opened 2025-04-11
Rewrite templates/partials for Hugo v0.146.0

ragibson:
Blowfish's usage of `{{ partial "partials/..." }}` and `{{ if templates.Exists "partials/..." }}` no longer exist in Hugo v0.146.0 since they reworked all the templates logic (e.g., see https://github.com/gohugoio/hugo/issues/13553).

I've replaced those constructs with the recommendations in the new documentation (e.g., https://gohugo.io/functions/templates/exists/).

This should fix https://github.com/nunocoracao/blowfish/issues/2053 and should fix https://github.com/nunocoracao/blowfish/issues/2052.

Issue #2053, opened 2025-04-10
[bug] Site fails to build on Hugo 0.146


Ast3risk-ops:
**Describe the bug**
It displays this error:
```
Error: html/template:_shortcodes/figure.html:3:14: no such template "_internal/shortcodes/figure.html"
```

I imagine this is due to the new templating system introduced in Hugo 0.146.

**To Reproduce**
1. try to build with Hugo 0.146

**Expected behavior**
It works

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux

**Hugo & Blowfish versions**
`0.146.1+extended+withdeploy`
latest Blowfish
**Additional context**

eallion:
same

rroethof:
same

rroethof:
It seems to be impacted by https://github.com/gohugoio/hugo/issues/13553

mer0x:
same

ragibson:
Took a shot at fixing this in https://github.com/nunocoracao/blowfish/pull/2055

iunn-sh:
same

nunocoracao:
Fixed

Issue #2052, opened 2025-04-10
No figure shortcode in newer versions of Hugo


bretton:
**Describe the bug**
When running `hugo -D` or `hugo` with homebrew-installed version on a mac I get the following error:
```
Error: html/template:_shortcodes/figure.html:3:14: no such template "_internal/shortcodes/figure.html"
```

it seems to be from 
```
themes/blowfish/layouts/shortcodes/figure.html:  {{ template "_internal/shortcodes/figure.html" . }}
```

**To Reproduce**
In a new hugo site (steps copied from an internal freebsd pot image script)
```
git submodule add --depth=1 https://github.com/nunocoracao/blowfish.git themes/blowfish
```
Then do extra steps to add site content, then build with `hugo -D` or `hugo`

**Expected behavior**
Site is built

**Desktop (please complete the following information):**
 - OS: Macos, brew install for hugo

**Hugo & Blowfish versions**
hugo v0.146.1+extended+withdeploy darwin/arm64 BuildDate=2025-04-10T16:44:46Z VendorInfo=brew

blowfish latest from repo

**Additional context**
I'm using an earlier hugo in live setup and it works (hugo v0.140.2+extended freebsd/amd64 BuildDate=2024-12-30T17:44:54Z+0000 VendorInfo=freebsd)

but it's not working when trying to test with newer version of hugo installed via homebrew on a mac. 

Noting it here as it may need to be sorted out when my live setup has an upgraded hugo version.

DarkXero-dev:
I can confirm this issue on Arch Linux

eallion:
Same on Win11 Scoop.

ragibson:
Took a shot at fixing this in https://github.com/nunocoracao/blowfish/pull/2055

eclecticpassions:
I haven't touched my site lately, but today I ran it and I got the same error after updating to `hugo v0.147.0+extended+withdeploy`. I don't know if this works for everyone, but I got the errors to go away and can run `hugo -D serve` with no errors.

1. Create a `figure.html` in the root of your project under `yoursite/layouts/shortcodes/figure.html`. 
2. Go to Hugo's offical [GitHub repo](https://github.com/gohugoio/hugo), search for `figure.html` and copy the contents and paste it into your shortcode created in step 1.

![Image](https://github.com/user-attachments/assets/71b954c5-d872-4267-b676-e1157c67129f)

[...]

PR #2210, opened 2025-06-05
🐛 fix: add default figure shortcode as partial

f-hollow:
I updated the blowfish submodule in my Hugo project and ran into the error `partial "shortcodes/figure.html" not found`. It has already been discussed in #2052, #2053, then reported again in #2093. However, [this fix](https://github.com/nunocoracao/blowfish/pull/2055/files#diff-dd648b847f1ddc24890196195d38087229de05e7b6a08bbf2de0da28c4765244R3) from #2053 doesn't solve the problem, because, from what I gather, `partial` doesn't have access to embedded shortcodes anymore.

This PR has the embedded `figure` shortcode renamed and copied to partials. This seems to be the way to fix such issues as of now.

**Update**:

I see that another way to fix this is to revert the fix linked above for the figure shortcode and keep `{{ template "_internal/shortcodes/figure.html" . }}`. Even though, it is [discouraged](https://github.com/gohugoio/hugo/issues/13553) by Hugo maintainers. But, in this case, the existing `figure.html` shortcode with the custom logic should be renamed, which, in turn, means that the shortcode names for non-`default-true` figures in all existing content would need to be updated accordingly. So the proposed fix still seems like a safer alternative. It is also better in terms of maintaining the existing content.

> [!WARNING] 
> This PR breaks the project's [coding convention](https://github.com/nunocoracao/blowfish/blob/main/CONTRIBUTING.md#coding-conventions): `Avoid code reuse in templates by extracting components into partials.`
>
> However, since Hugo v0.146 removed access to embedded templates like "_internal/shortcodes/figure.html", I see no other way to deal with the issue, except for copying the Hugo's embedded shortcode into `layouts/partials/hugo-embedded/shortcodes/figure-default.html`.

### DIFF
--- a/layouts/partials/hugo-embedded/shortcodes/figure-default.html
+++ b/layouts/partials/hugo-embedded/shortcodes/figure-default.html
@@ -1,0 +1,43 @@
+{{/*
+  Copied from Hugo v0.146
+  Source: https://github.com/gohugoio/hugo/blob/83cfdd78ca6469e6d7265323d9fad1448880e559/tpl/tplimpl/embedded/templates/_shortcodes/figure.html
+*/}}
+
+<figure{{ with .Get "class" }} class="{{ . }}"{{ end }}>
+  {{- if .Get "link" -}}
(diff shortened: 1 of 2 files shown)


f-hollow:
@nunocoracao PTAL

PR #2211, opened 2025-06-05
:recycle: refactor: Move most inline styles out of line

servedsmart:
- Move out of line by using already defined css classes
- Move paddings out of line
- Move margins out of line
- Move z-indexes out of line
- Move center-relative-left out of line

I haven't tested everything in detail, but after a few minutes of navigating I haven't noticed any differences.

I'm not sure if I'm supposed to build `main.css` manually or if that is done while deploying. If it isn't please let me know and I will rebuild it. (Solved: https://github.com/nunocoracao/blowfish/pull/2211#issuecomment-2945422610)

The reason for this is also related to [this discussion](https://github.com/nunocoracao/blowfish/discussions/2198). It however should also improve readability and especially maintainability. Also everything being in `main.css` as far as I know could even improve performance over inline styles but I'm not sure.

(Showcase: https://github.com/nunocoracao/blowfish/pull/2211#issuecomment-2980420506)

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -3698,6 +3698,42 @@ pre {
   height: 0px;
   visibility: hidden;
 }
+.center-relative-left {
+  left: calc(max(-50vw,-800px) + 50%);
+}
+.margin-0 {
+  margin: 0;
+}
+.margin-top-\[-15px\] {
+  margin-top: -15px;
+}
+.margin-top-\[0\.5rem\] {
+  margin-top: 0.5rem;
+}
+.margin-right-\[10px\] {
+  margin-right: 10px;
+}
+.margin-left-\[0px\] {
+  margin-left: 0px;
+}
+.padding-main-menu {
+  padding: 2px 0 3px 0;
+}
+.padding-top-\[5px\] {
+  padding-top: 5px;
+}
+.z-index-\[-10\] {
+  z-index: -10;
+}
+.z-index-80 {
+  z-index: 80;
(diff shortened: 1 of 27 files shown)


servedsmart:
Just as a sidenote: I didn't use the same css classes as tailwind generates because I didn't want to cause any conflicts with classes that tailwind might add (fex. `z-100/`). If I should stick to the exact naming scheme as in the generated `main.css` please let me know!

This should btw decrease the size of all inline style hashes by a lot. I will later add a link to the discussion with the resulting Cloudflare API HTTP request. (later might also be tomorrow+).

servedsmart:
Okay I just did a few more minutes of testing. There is still some errors to fix and work to do. Sorry for my initial optimism.

servedsmart:
> I'm not sure if I'm supposed to build `main.css` manually or if that is done while deploying. If it isn't please let me know and I will rebuild it.

It seems like I should have definetly rebuilt it. I have just done that. This should probably fix the bugs I have encountered on the deployment preview.

servedsmart:
Yes, the bugs I have encountered were just because I haven't rebuilt `main.css`

servedsmart:
This is a small showcase of https://github.com/nunocoracao/blowfish/commit/d97b1e5c38b0f66412a8ab89f717f177ff9d47b5:

https://github.com/user-attachments/assets/249300d0-c35e-4ea9-9bd7-ee5393774380

Not all changes are included here, but I'm pretty sure that this should be enough.

PR #2218, opened 2025-06-11
:recycle: refactor: Move all inline event handlers out of line

servedsmart:
- Added .js files are included in `main.bundle.js`

Not using inline event handlers is [encouraged](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Events#inline_event_handlers_%E2%80%94_dont_use_these).
I am not 100% sure if I have caught all inline event handlers but I couldn't find any by searching for terms in [this list](https://www.w3schools.com/tags/ref_eventattributes.asp) (not the official spec but I couldn't find anything that lists every possible attribute).

For merging this into dev with changes from my recent unmerged prs, I will still have to make some changes. I just didn't want to include changes from these prs here.

(Showcase: https://github.com/nunocoracao/blowfish/pull/2218#issuecomment-2980190721):

### DIFF
--- a/layouts/partials/article-link/card-related.html
+++ b/layouts/partials/article-link/card-related.html
@@ -1,9 +1,9 @@
 {{ $disableImageOptimization := .Page.Site.Params.disableImageOptimization | default false }}
 <div class="min-w-full min-h-full border border-neutral-200 dark:border-neutral-700 border-2 rounded overflow-hidden shadow-2xl relative">
   {{ with .Params.externalUrl }}
-    <a href="{{ . }}" target="_blank" rel="external" class="absolute inset-0"></a>
+    <a href="{{ . }}" target="_blank" rel="external" class="absolute inset-0" aria-label="{{ $.Title }}"></a>
   {{ else }}
-    <a href="{{ .RelPermalink }}" class="absolute inset-0"></a>
+    <a href="{{ .RelPermalink }}" class="absolute inset-0" aria-label="{{ $.Title }}"></a>
   {{ end }}
   {{- with $.Params.images -}}
     {{- range first 6 . }}
--- a/layouts/partials/article-link/card.html
+++ b/layouts/partials/article-link/card.html
@@ -1,9 +1,9 @@
 {{ $disableImageOptimization := .Page.Site.Params.disableImageOptimization | default false }}
 <div class="min-w-full min-h-full border border-neutral-200 dark:border-neutral-700 border-2 rounded overflow-hidden shadow-2xl relative">
   {{ with .Params.externalUrl }}
-    <a href="{{ . }}" target="_blank" rel="external" class="absolute inset-0"></a>
+    <a href="{{ . }}" target="_blank" rel="external" class="absolute inset-0" aria-label="{{ $.Title }}"></a>
   {{ else }}
-    <a href="{{ .RelPermalink }}" class="absolute inset-0"></a>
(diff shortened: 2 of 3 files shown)


servedsmart:
I have tested the second commit quite extensively. All functionality for shortcodes works the same as it did before, I think that the added commit greatly improves readability and adds a lot of `resources.Minify | resources.Fingerprint`. I added it here since I modified the same file here.

servedsmart:
The latest commit has changed a lot of stuff, but after somewhat extensive testing of a lot of sites in the exampleSite I am almost certain that it works just the same with these improvements that are also mentioned in the commit:

- This removes one script and is an easier and more standard way to handle opening links
- This also improves UX since hrefs will finally be visible in the browser. This as far as I know, is not possible with the previous approach
- I have also removed an href from a div, which as far as I know is invalid

The following video is just to showcase https://github.com/nunocoracao/blowfish/pull/2218/commits/29af6ec60242577a8be1736a32f9ee79d9004bdc (Superseded by: https://github.com/nunocoracao/blowfish/pull/2218#issuecomment-2980190721):

https://github.com/user-attachments/assets/60cc6e63-f00c-4639-8947-0c978b8aed06

servedsmart:
This is after applying https://github.com/nunocoracao/blowfish/commit/4fefcf305f1aeed83eea972ed116d52d5005738d which fixed an issue with `simple.html` and showcases the changes to cards:

https://github.com/user-attachments/assets/efb3aec3-198e-4b3c-bfb0-98d33a26d4fe

This is also with https://github.com/nunocoracao/blowfish/commit/4fefcf305f1aeed83eea972ed116d52d5005738d applied but just showcasing the general changes:

https://github.com/user-attachments/assets/e7c5154e-eaf1-4d4f-a842-e0f72c3e60d2

This is also with https://github.com/nunocoracao/blowfish/commit/4fefcf305f1aeed83eea972ed116d52d5005738d applied but just showcasing the like button (The javascript error is unrelated to my changes and expected since I don't have firebase configured as far as I know):

https://github.com/user-attachments/assets/d4aa0ece-bc9e-46ee-921f-7ff45a5cf182

servedsmart:
Commit https://github.com/nunocoracao/blowfish/pull/2218/commits/aaa07734fc8e2ebf77eb51dcbc1fc50401a5df81 works just the same as in the mentioned video. It#s just to avoid using too large z-index.

Commit https://github.com/nunocoracao/blowfish/commit/4fefcf305f1aeed83eea972ed116d52d5005738d also produces a few more lighthouse warnings. The same accessibility problems would have been there before, therefore this is a good thing in my opinion.

</img>

servedsmart:
This is a similar lighthouse report after adding the `aria-label` in https://github.com/nunocoracao/blowfish/pull/2218/commits/3dce0c6125d694086ac9848ec4db2a74e3ac14ab:

</img>

I don't think that `Touch targets do not have sufficient size or spacing.` is that relevant and it would require a lot of restyling. I could try to submit a new pr in the future for that though. Also it isn't new, it's just that lighthouse didn't detect it before. The anchor tags definitely help a lot with automated processing instead of using a custom javascript approach.

servedsmart:
@nunocoracao Hey, thanks for merging all those recent PRs so quickly. I'm pretty sure that you might've closed all the unmerged ones by deleting the `dev` branch. If not, just ignore this :)

Issue #1422, opened 2024-04-17
✨ Ads integration for posts.


harisheoran:
**Is your feature request related to a problem? Please describe.**
I am using a blowfish website for my blogs and it is absolutely Amazing, but to monetize the blogs it needs ads integration.

**Describe the solution you'd like**
an easy integration of ads service like google adsense or any popular global ad service.

nunocoracao:
Hello @harisheoran which ad services would you like to see integrated?

harisheoran:
Google adsense

PR #2224, opened 2025-06-16
✨ support advertisement

Weaxs:
add advertisement config

- support google adsense

https://support.google.com/adsense/answer/12169212

https://github.com/nunocoracao/blowfish/issues/1422


nunocoracao:
Thanks @Weaxs could you please add docs for this when you have time?

PR #2227, opened 2025-06-16
:recycle: Refactor: Improve consistency in <script> tags

servedsmart:
- Minify `zoom.min.js` with `resources.Minify`: I do get that one doesn't have to minify a script that already is, but just having this for consistency is in my opinion a good idea. It also reduces the size of `zoom.min.js` from `9637 bytes` to `9510 bytes`.
- Fix inconsistencies in other `<script>` elements

(Showcase: https://github.com/nunocoracao/blowfish/pull/2227#issuecomment-2980367849)

PR #2232, opened 2025-06-17
:recycle: Refactor: Improve consistency in <script> tags

servedsmart:
- Add `type="text/javascript"` and restructure attributes in various tags
- Some scripts have been left out because they are included in https://github.com/nunocoracao/blowfish/pull/2231

I am resubmitting this PR with required changes to merge into current main.

This PR was created due to a possibly unwanted closure of [my previous PR](https://github.com/nunocoracao/blowfish/pull/2227). See [this comment](https://github.com/nunocoracao/blowfish/pull/2218#issuecomment-2982008992) for details.  Sorry if you didn't want me to do that again just to be sure.

PR has been confirmed working after 1min of testing :)

EDIT: I am no longer using `resources.Minify` since that might've mostly been a personal preference and I don't necessarily want to get into a disagreement.

## Purpose

- Right now, we sometimes use `type="text/javascript"` and in other cases omit it. These changes improve consistency, readability and maintainability since people will know right away what type of `<script>` they are looking at and will no longer stumple upon inconsistencies in the codebase.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -44,13 +44,14 @@
   </body>
   {{ if .Site.Params.buymeacoffee.globalWidget | default false }}
     <script
-      data-name="BMC-Widget"
-      data-cfasync="false"
+      type="text/javascript"
       src="https://cdnjs.buymeacoffee.com/1.0.0/widget.prod.min.js"
-      data-id="{{ .Site.Params.buymeacoffee.identifier }}"
-      data-description="Support me on Buy me a coffee!"
-      data-message="{{ .Site.Params.buymeacoffee.globalWidgetMessage | default "" }}"
+      data-cfasync="false"
       data-color="{{ .Site.Params.buymeacoffee.globalWidgetColor | default "#FFDD00" }}"
+      data-description="Support me on Buy me a coffee!"
+      data-id="{{ .Site.Params.buymeacoffee.identifier }}"
+      data-message="{{ .Site.Params.buymeacoffee.globalWidgetMessage | default "" }}"
+      data-name="BMC-Widget"
       data-position="{{ .Site.Params.buymeacoffee.globalWidgetPosition | default "Left" }}"
       data-x_margin="18"
       data-y_margin="18"></script>
--- a/layouts/partials/analytics/fathom.html
+++ b/layouts/partials/analytics/fathom.html
@@ -1,11 +1,13 @@
 {{ if isset site.Params.fathomAnalytics "domain" }}
   <script
     defer
+    type="text/javascript"
     src="https://{{ site.Params.fathomAnalytics.domain }}/script.js"
     data-site="{{ site.Params.fathomAnalytics.site }}"></script>
 {{ else }}
   <script
     defer
+    type="text/javascript"
     src="https://cdn.usefathom.com/script.js"
     data-site="{{ site.Params.fathomAnalytics.site }}"></script>
 {{ end }}
--- a/layouts/partials/analytics/ga.html
+++ b/layouts/partials/analytics/ga.html
@@ -1,7 +1,8 @@
 <script
   async
+  type="text/javascript"
   src="https://www.googletagmanager.com/gtag/js?id={{ site.Config.Services.GoogleAnalytics.ID }}"></script>
-<script>
+<script type="text/javascript">
   window.dataLayer = window.dataLayer || [];
   function gtag(){dataLayer.push(arguments);}
   gtag('js', new Date());
--- a/layouts/partials/analytics/seline.html
+++ b/layouts/partials/analytics/seline.html
@@ -1,5 +1,6 @@
 <script
   async
+  type="text/javascript"
   src="https://cdn.seline.so/seline.js"
   data-token="{{ site.Params.selineAnalytics.token }}"
   data-id="seline-script"></script>
--- a/layouts/partials/analytics/umami.html
+++ b/layouts/partials/analytics/umami.html
@@ -1,19 +1,21 @@
 {{ if isset site.Params.umamiAnalytics "domain" }}
   <script
-    data-id="umami-script"
     async
+    type="text/javascript"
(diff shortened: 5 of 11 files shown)


ZhenShuo2021:
Hi, regarding this specific PR, the attribute `type="text/javascript"` [can be omitted](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type) in modern HTML as it's the default.

From some of your recent PRs, I've noticed a few patterns that might help with the review process:

1. It would be great if PR changes could match exactly what's described in the title and message - this helps owner know exactly what to expect
2. Keeping each PR small and atomic help owner focus on one clear purpose at a time  
3. It might be helpful to explain the practical benefits of the changes to help the owner better understand their value

Hope this helps!

servedsmart:
> From some of your recent PRs, I've noticed a few patterns that might help with the review process:
> 
>     1. It would be great if PR changes could match exactly what's described in the title and message - this helps owner know exactly what to expect
> 
>     2. Keeping each PR small and atomic help owner focus on one clear purpose at a time
> 
>     3. It might be helpful to explain the practical benefits of the changes to help the owner better understand their value
> 
> 
> Hope this helps!

Thank you very much for the suggestions. I will try to keep this in mind in the future.

Regarding point 1, I didn't want to have titles that are way too long, therefore it is quite generalized.

If you could maybe elaborate on point 2: I'm pretty sure that is what I've done most of the time, but please let me know why you don't think so. EDIT: I found one example - the changes to `vendor.html` in https://github.com/nunocoracao/blowfish/pull/2231 . But I'm pretty sure that other recent PRs were quite focused on one objective.

And point 3 is definetly valid, I will try to elaborate more in future PRs. :+1:

EDIT: I have added changes to all currently open PRs to address your comment. Thank you very much again for pointing this out!

servedsmart:
> Hi, regarding this specific PR, the attribute `type="text/javascript"` [can be omitted](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type) in modern HTML as it's the default.

I know that, but to explain further:

Right now, the codebase sometimes uses the `type=text/javascript` attribute and sometimes doesn't. It being there and consistent does definetly provide a benefit since people will know right away why it is there and what the `<script>` tag actually is. Also consistency does in any case (if not for sure almost any cases) improve readability and maintainability.

ZhenShuo2021:
> If you could maybe elaborate on point 2....

TL;DR: Clean, small and exact PRs are preferred since reviewers have limited time to go through lengthy descriptions.

---

Thanks for the response! Let me clarify what I meant by point 2.

For this current PR, you mention "Some scripts have been left out because they are included in #2231". This makes it non-atomic since it depends on another PR's scope. A more atomic approach might be: add all `type="text/javascript"` attributes in one PR, then create a separate follow-up PR based on that one.

When someone first sees a description like this, the immediate reaction is often "this looks like a mixed change" and they might not dig deeper to understand the reasoning.

Looking at some other examples:

- In #2231 "Move all inline event handlers out of line", there are minify code changes and aria-label additions that don't quite match the title
- In #2249 "Refactor: Autogen classes and use consistent naming", the actual changes seem to be fixing incorrect Tailwind usage rather than refactoring. These syntactically incorrect classes should have no effect at all, so the proper fix would probably be to remove them. Since no one has raised concerns about the styling, it suggests people are satisfied with the current appearance. If you "fix" the syntax, this PR would essentially become a styling change.
- On a related note, large PRs with extensive commit history plus cross-references to other PRs means reviewers need to process quite a bit of information to understand the full scope.

For smaller changes, you often don't need lengthy descriptions. Title + diff can be self-explanatory. Too much text can sometimes make it harder to find the key points. Personally, I prefer a few clear lines about the benefits, with additional explanation only for parts the code can't explain itself.

Hope this helps clarify things!

ZhenShuo2021:
> Right now, the codebase sometimes uses the `type=text/javascript` attribute and sometimes doesn't...

The attribute doesn't actually cause confusion. Personally, I'd prefer to leave it as it was before. If consistency is really needed, I think removing all `type=text/javascript` attributes would be a better choice. This makes fewer changes and follows MDN guidelines.

servedsmart:
Thank you for elaborating on your previous comment.

> "Refactor: Autogen classes and use consistent naming", the actual changes seem to be fixing incorrect Tailwind usage rather than refactoring. These syntactically incorrect classes should have no effect at all [...]

- Please see [the diff to the compiled main.css](https://github.com/nunocoracao/blowfish/pull/2249/files). This isn't a style change, it just refactors by changing the names of classes used in html. These classes actually had an effect because they were also in the `css`. I'm pretty sure that the title/explaination is quite accurate in that example.

Regarding your first example: I will rename https://github.com/nunocoracao/blowfish/pull/2231 to not include refactor in the title, but submitting a new pr just for adding accessibility features I think would be a lot of effort if you can just add it in the same PR.

I get your comments about complexity and I am sorry for "spamming" 😆 commits and force pushing that often. I'm not used to collaborative projects at all and at least most of my past have just coded for myself or submitted small changes. Thank you very much for your feedback though and I will keep all of your points in mind in the future.

EDIT: I will also close this PR and maybe resubmit it later. I do think that having `type="javascript"` attributes is useful. Before submitting a new PR for that, I will make a poll in discussions.

ZhenShuo2021:
These are just my personal thoughts, but I believe atomic commits are generally preferred because this is how Git is designed.

Also, I think it’s perfectly fine to create a PR directly. It’s more efficient and the collaboration workflows differ depending on the project owner.

servedsmart:
> These are just my personal thoughts, but I believe atomic commits are generally preferred because this is how Git is designed.
> 
> Also, I think it’s perfectly fine to create a PR directly. It’s more efficient and the collaboration workflows differ depending on the project owner.

Yeah, you are correct. I mean ideally I would also prefer that. I just split up the https://github.com/nunocoracao/blowfish/pull/2231 into 3 commits to at least somehow follow your feedback. It was just way easier for my workflow to do it as I did before, therefore the not so good PRs. I understand that this does create issues for people reviewing the PRs.
Thanks again for your help! I hope that I can pay more attention in future commits :+1:

PR #2245, opened 2025-06-18
♻️ Refactor: make article-link formattable

ZhenShuo2021:
This PR abstracts anchor attributes to enable formatting of article links.

To achieve this, an additional partial file was introduced. Given that only three files require formatting, the cost of adding a new file should be carefully evaluated. While using an [inline partial](https://gohugo.io/functions/go-template/define/) can avoid creating a new file, embedding an inline partial inside the article-link file may cause confusion for users.

This PR performs the same task as #2179.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -28,8 +28,8 @@ static
 layouts/_default/_markup/render-heading.html
 layouts/_default/index.json
 layouts/_default/list.html
-layouts/partials/hero/background.html
 layouts/shortcodes/screenshot.html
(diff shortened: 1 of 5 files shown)


nunocoracao:
@ZhenShuo2021 conflicts - can you fix?

ZhenShuo2021:
@nunocoracao Fixed. Since some changes were covered in other PRs, I closed #2246 #2247 #2248  and merged the remaining updates into this one.

Issue #2256, opened 2025-06-19
Umami analysis has bug in v2.87.0


Weaxs:
**Describe the bug**
v2.87.0 pretty run commit

before:
```
src="https://{{ site.Params.umamiAnalytics.domain }}/{{ with site.Params.umamiAnalytics.scriptName }}{{ . }}{{ else }}script.js{{ end }}"
```
 after:
```
    src="https://{{ site.Params.umamiAnalytics.domain }}/{{ with site.Params.umamiAnalytics.scriptName }}
      {{ . }}
    {{ else }}
      script.js
    {{ end }}"
```

actual:
```
  <script data-id="umami-script" async src="https://umami.weaxsey.org/
      script.js
    " data-website-id="" data-domains="" type="a7fac8f00ef64e85e21fc4a7-text/javascript"></script>

```

actual request；

https://umami.weaxsey.org/%20%20%20%20%20%20script.js

![Image](https://github.com/user-attachments/assets/b2fc2944-b383-4219-923f-fd1befbfe7d4)

https://github.com/nunocoracao/blowfish/commit/f2e224a042f273a2cef7d46cb9e5634f38844c1f#diff-d7aed1e2ba4542775b6f38c4be562012be7430b6ed2463dd577a59190b532722

**To Reproduce**

**Hugo & Blowfish versions**
v2.87.0

PR #2257, opened 2025-06-19
umami bugfix

Weaxs:
https://github.com/nunocoracao/blowfish/issues/2256


ZhenShuo2021:
These can be added to prevent the code from being formatted again:

```
{{/* prettier-ignore-start */}}
...
{{/* prettier-ignore-end */}}
```

PR #2260, opened 2025-06-19
:lock: Use pure version of medium-zoom to avoid inline style and simplify CSP usage

servedsmart:
This has been tested for if it creates any inline styles (By deploying it to a site with a CSP that doesn't hash one for medium-zoom) and for functionality and is working flawlessly.

## Purpose

This avoids an extra sha sum for a hashed CSP because it merges the CSS into main.bundle.css and avoids any inline styles being applied by medium-zoom.

This addresses concerns mentioned in the closed PR https://github.com/nunocoracao/blowfish/pull/2196, now the same changes are integrated into package.json.

This fixes https://github.com/nunocoracao/blowfish/issues/2195.


servedsmart:

ZhenShuo2021:
@servedsmart Hi, I'm not a professional frontend developer, so I might be wrong, but I have concerns about this CSP fix.

According to the [medium-zoom documentation](https://github.com/francoischalifour/medium-zoom/blob/35b519fbdef97df69cc1d7daf46988effef0860b/README.md?plain=1#L131-L137), the pure version is designed to "control when to inject the Medium Zoom CSS styles" - not avoid injecting them. The [original PR #206](https://github.com/francoischalifour/medium-zoom/pull/206) was for Astro Page Transitions support.

This suggests the pure version still performs dynamic CSS injection, which would trigger CSP violations even with CSS pre-loaded in the main bundle.

Could you clarify if you tested this with a CSP that blocks `unsafe-inline` styles?

Thanks!

servedsmart:
[This](https://gist.github.com/servedsmart/33fa662670e0520315cc0e979c3dfd84) is a request containing my CSP. At that time and also right now I use this commit in my own site. So I can definetly confirm that this is working.

EDIT: I accidentaly have `unsafe-inline` on my current site for `style-src`. Forgot to remove it while debugging, I will provide a working current example in a few hours.

Also see https://github.com/nunocoracao/blowfish/discussions/2198 for some more context on all of this.

To address your concern: Their documentation is somewhat unclear. If you look at the javascript file, you can see that no dynamic inline css is added. At least that's what I have observed when looking at it.
I'm pretty sure that by "when to inject it", they mean when to load it from either javascript/html or whatever.

ZhenShuo2021:
Thanks for the update.

This doesn't address the core issue, the [zoom](https://github.com/francoischalifour/medium-zoom/blob/35b519fbdef97df69cc1d7daf46988effef0860b/src/medium-zoom.js) relies on inline styles like transform, so `unsafe-inline` is still required.

Furthermore, even if this approach did work, it relies on undocumented behavior, not a stated feature. Relying on such implementation details is fragile. We should stick to documented features only.

servedsmart:
https://gist.github.com/servedsmart/df56779a01182f4bb3f3fd1cdb657180 This is the current request. It works without any problems and is not using `unsafe-inline`. Also see https://autoshop.servedsmart.top and its' headers to see this working.

Also see https://github.com/nunocoracao/blowfish/discussions/2198#discussioncomment-13728194 for more details.

EDIT: The transform is covered by 'strict-dynamic' in script-src which is a lot safer than unsafe-inline.

servedsmart:
Also I would disagree, the quote you have also used:

> If you want to control when to inject the Medium Zoom CSS styles, you can use the pure JavaScript bundle:

Means that one can use the bundle. It doesn't mean that it has to be an import or that it has to be loaded via javascript just because the code example uses that. At least that's what I''d understand.

What the previous script did was define css inline, just as if it would be a file and then load that dynamically, this is very hard to generate a hash for and therefore almost impossible to safely implement in a CSP.

EDIT: [This](https://diffy.org/diff/b854dd5a5c56e) is the diff of changes added by the previous implementation (test2.js is the old implementation, test1.js the current one in main). The only real difference is that there is no styleSheet created within javascript inline code.
The diff is on an external site since I don't know how to share it on github.

ZhenShuo2021:
I'd appreciate if you could double check the source code, since transform is clearly used there.

servedsmart:
> I'd appreciate if you could double check the source code, since transform is clearly used there.

I didn't say that it isn't, just that it works a lot better than the old approach for implementing a CSP.

Just as a sidenote:

The `'unsafe-inline'` in script-src is ignored because there are hashes specified (This is done for backwards compatibility).
[securityheaders.com](https://securityheaders.com/?q=https%3A%2F%2Fautoshop.servedsmart.top&followRedirects=on) gives A+ for the site I have used as an example.

ZhenShuo2021:
Hi @servedsmart,

I want to acknowledge that I was wrong in my assessment regarding the CSP behavior. I realize that my original assumption about the inline styles was incorrect. I appreciate that you took the time to explain this thoroughly and provide concrete examples.

My previous comment on this topic was dismissive and I regret how that came across. I should have been more careful in how I expressed disagreement, especially since you were providing data to support your point.

Thanks for your patience in following through with the details.

servedsmart:
> Hi @servedsmart,
> [...]

Hi @ZhenShuo2021 ,

that is totally fine. My original description of this PR was lacking that data and details, therefore I understand your original skepticism.
I also understand that I probably lost some trust because of the mistakes in earlier PRs.

Thank you very much for clarifying and acknowledging the mistakes in your assumption, that gives me confidence that we can continue our work on blowfish in a similar constructive manner.

As I've said before, I will try my best to take more care in future PRs and to not commit anything that will be a burden for other maintainers. That should make collaboration a lot easier. (For the next few days I don't have much time, but I'm happy to come back later).

Btw thank you for your awesome work, quick fixes and all that you've done for the blowfish theme way before I made my first commit.🥇

PR #2261, opened 2025-06-20
:adhesive_bandage: Fix: Incorrect usage of <label> for menu-button

servedsmart:
This has been tested and works just like using a `<label>` did before, also on different screen sizes. Chrome no longer shows an issue for this.

## Purpose

This fixes the following issue in Chrome: No label associated with a form field.

I don't know why a <label> was used here before, but I see no reason to do so, a div works just as well.


servedsmart:

PR #2265, opened 2025-06-20
🤖 Fix JS issues and modernize node scripts

nunocoracao:
## Summary
- allow likes to toggle via `let` instead of `const`
- drop debug logging from search
- read Packery options from HTML data attributes
- convert node scripts to async fs usage and `const`/`let`

## Testing
- `npm test` *(fails: Missing script)*

------
https://chatgpt.com/codex/tasks/task_e_6851e9cb5e8483269c5743037a5af682


servedsmart:
Thank you for fixing the `const liked_page`, it seems like I haven't tested this properly. I apologize for that.
This should fix the later reported https://github.com/nunocoracao/blowfish/issues/2266

PR #2269, opened 2025-06-22
:recycle: Refactor Use universal json and move styles out of line

servedsmart:
This has been tested for 10 minutes in different scenarios and is working just like it did before.

## Purpose

- Having 4 json files with different names but the same content isn't a good idea.
- Using classes here is quite straight-forward and reduces the amount of inline styles if people use any of the relevant repo shortcodes. This can reduce the size of a CSP and improve maintainability by not having the same inline script logic in 4 different files.


servedsmart:
Converted to draft to find a way to save bytes because this adds a lot to `main.css`. (`37119` bytes)

EDIT: This has been resolved in https://github.com/nunocoracao/blowfish/pull/2269/commits/9ac693e971db04d19ec05967457e446172a0fb12. Now only relevant classes will be created and for `exampleSite` we only add `228` bytes

nunocoracao:
@servedsmart  conflicts - can you fix?

servedsmart:
> @servedsmart conflicts - can you fix?

All conflicts are resolved now.

servedsmart:

PR #2274, opened 2025-06-23
feat(gallery): support images from static folder

ZhenShuo2021:


ZhenShuo2021:
Close this PR because it already support images from static folder.

PR #2277, opened 2025-06-24
:sparkles: Add leaflet shortcode that adds map without storing cookies

servedsmart:
This has been tested for functionality and every shortcode parameter has been tested separately.

## Purpose

This does not store any cookies/localstorage etc. and therefore should be easy to implement even for EU countries.

As far as I know, using the iframe from [OpenStreetMap](https://www.openstreetmap.org/) or [UMAP](https://umap.openstreetmap.de/) is not compatibly with a lot of security headers and requires storing a cookie. Because of the cookie, it requires `Cross-Origin-Embedder-Policy: unsafe-none` according to my testing.

This is compatible with the following HTTP security headers (only relevant ones are included):

```
img-src 'self' blob: data: https://tile.openstreetmap.org
Cross-Origin-Embedder-Policy: credentialless
Cross-Origin-Resource-Policy: same-origin
```

## Example

[This](https://deploy-preview-2277--snazzy-dango-efb2ec.netlify.app/docs/shortcodes/#leaflet) is the documentation with an example of how the map would look.

### DIFF
--- a/assets/js/leaflet.js
+++ b/assets/js/leaflet.js
@@ -24,7 +24,9 @@
   const mapWidth = script?.getAttribute("data-map-width");
 
   const mapElement = document.getElementById(mapId);
+  mapElement.style.maxHeight = "100vh";
   mapElement.style.height = mapHeight;
+  mapElement.style.maxWidth = "80vw";
   mapElement.style.width = mapWidth;
 
   const map = L.map(mapId).setView([mapCenterX, mapCenterY], defaultZoom);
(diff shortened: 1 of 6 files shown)


servedsmart:
There is currently an issue with z-indexes that are being assigned by leaflet not being compatible with the theme layout, therefore this needs some more work and testing.

servedsmart:
I will switch this to OpenLayers for myself. I might submit a PR for that later. It isn't as straight forward as this, therefore it might be hard to maintain here.

If someone would want that feature though, I will definetly add a PR here. I have got it working quite well at the moment but my current implementation is still being tested.

PR #2278, opened 2025-06-24
:refactor: Use IIFE for background-blur and simplify code

servedsmart:
This is inspired by https://github.com/nunocoracao/blowfish/pull/2262 . If both are merged, this would enhance consistency, if not this is still useful.

## Purpose

This improves readability and removes an unnecessary function declaration by using IIFE.


servedsmart:
All commits that have been made in the last 20mins in all my open PRs were due to a rebase.

PR #2338, opened 2025-07-16
🔧 Chore: remove prettier-plugin-tailwind

ZhenShuo2021:
The behavior is inconsistent in dark mode and with some hover interactions. Generally, we prefer to order classes by shape, margin, and padding first, followed by color, and finally by variants such as dark mode and hover. This is also how Blowfish was originally structured.

However, even though the [documentation](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) [states this](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted), it formats the code like this:

```diff
-  class="m-1 rounded bg-neutral-300 p-1.5 text-neutral-700 hover:bg-primary-500 hover:text-neutral dark:bg-neutral-700 dark:text-neutral-300 dark:hover:bg-primary-400 dark:hover:text-neutral-800"
+  class="hover:bg-primary-500 hover:text-neutral dark:hover:bg-primary-400 m-1 rounded bg-neutral-300 p-1.5 text-neutral-700 dark:bg-neutral-700 dark:text-neutral-300 dark:hover:text-neutral-800"
```

Since the current setup works well, there is no need to introduce another tool that causes more issues than the value it provides.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,8 +7,7 @@
   "proseWrap": "always",
   "plugins": [
     "@awmottaz/prettier-plugin-void-html",
-    "prettier-plugin-go-template",
-    "prettier-plugin-tailwindcss"
+    "prettier-plugin-go-template"
   ],
   "overrides": [
     {
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "packery": "^3.0.0",
     "prettier": "^3.6.2",
     "prettier-plugin-go-template": "^0.0.15",
-    "prettier-plugin-tailwindcss": "^0.6.13",
     "puppeteer": "^24.12.1",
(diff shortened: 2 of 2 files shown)


nunocoracao:
@ZhenShuo2021 conflicts here can you take a look?
Also, not sure I get this one - is this breaking something i.e. why remove?

ZhenShuo2021:
This configuration was introduced in issue #2267 in version v2.88.0. However, the plugin has never been used before. Even the prettier run in commit f2e224a did not include this plugin.

I want to remove it because the plugin does not sort classes as claimed, as shown in my PR description.

Update: I think the problem is my PR description is not clear enough, I've updated the description.

PR #2340, opened 2025-07-16
Dynamic attributes on Button shortcode

### LINKED ISSUE #2339: Allow for more attributes in buttons
It's frustrating trying to write hacky code in articles when I need to use a button for something other than simple link opening.

Use dynamic attributes on button shortcodes instead

Re-creating buttons from scratch without using shortcode

hudsonm62:
Closes #2339

As it stands, only the 3 predefined attributes on button shortcodes can be used. This isn't enough for a lot of use cases, other than simple link opening

By using dynamic attributes, we aren't breaking existing buttons and allowing for more attributes to used such as `download=""`, `style=""`, etc.


nunocoracao:
Not sure I love this change. But I don't mind giving it a test.
@hudsonm62 would you mind updating the docs for the button shortcode so that I can check an example of this?

PR #2252, opened 2025-06-18
:children_crossing: UX: Use anchor tags in article-meta/basic.html

servedsmart:
- This removes an inline event handler
- This also adds an aria-label

## Purpose

Using javascript instead of just using `<a>` tag is bad practice regarding accessibility and automated processing of web pages. Even the browser doesn't show a hint if hovering badges when using the old implementation. This in my opinion is a good way to replace inline event handlers in that code.

## Notes

Previously, this was part of https://github.com/nunocoracao/blowfish/pull/2231 and https://github.com/nunocoracao/blowfish/pull/2218, but after some thinking, I just split up the PR.

## Warning

> :warning: This currently isn't compatible with https://github.com/nunocoracao/blowfish/pull/2249 and https://github.com/nunocoracao/blowfish/pull/2268. Please notify me to make the necessary changes.

PR #2297, opened 2025-06-30
✨ Feat: add options for article shortcode

ZhenShuo2021:
Introduce showSummary and compactSummary options for article shortcode. Resolves #2295.

The `article-link/_shortcode.html` partial is almost the same as the original `article-link/simple.html`, but the element padding is tweaked to make the card look good with every option.

A separate _shortcode.html file is necessary to avoid affecting the many files that rely on simple.html. Mixing shortcodes with regular template usage would make it harder to maintain, especially with the addition of extra options.

Please note that the layout issue referenced in #2295 was addressed in #2035.

PR #2351, opened 2025-07-17
:ambulance: Hotfix for issues in #2252

servedsmart:
This addresses all issues mentioned in: https://github.com/nunocoracao/blowfish/pull/2252#issuecomment-3084947371

@nunocoracao please merge this before publishing the next release. Thanks in advance!

:warning: There are probably merge conflicts with https://github.com/nunocoracao/blowfish/pull/2352 . To fix without me just merge this first and then merge only the change to `$target` from https://github.com/nunocoracao/blowfish/pull/2352

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -1,4 +1,4 @@
-/*! tailwindcss v4.1.8 | MIT License | https://tailwindcss.com */
+/*! tailwindcss v4.1.11 | MIT License | https://tailwindcss.com */
 /*! Blowfish | MIT License | https://github.com/nunocoracao/blowfish */
 @layer properties;
 #zen-mode-button {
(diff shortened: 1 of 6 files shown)


servedsmart:
One style change I have observed:

Now the title will always be underlined if hovering the card. I'm pretty sure that this is actually an improvement because the whole card is clickable anyways. EDIT: Minor issue fixed in https://github.com/nunocoracao/blowfish/pull/2351/commits/5b1a8c7a55f32570393ed128ff655b454e5869bc

nunocoracao:
@servedsmart @ZhenShuo2021 may I ask you for help solving these conflicts? I merged a CSS refactor before seeing this.

I wonder if the refactor solves the initial problem though

servedsmart:
> @servedsmart @ZhenShuo2021 may I ask you for help solving these conflicts? I merged a CSS refactor before seeing this.
> 
> I wonder if the refactor solves the initial problem though

I'd assume that you mean https://github.com/nunocoracao/blowfish/pull/2333 by "refactor". This would not have solved the issue.

The issues were:

- The title of cards (related etc.) does not get underlined, I did not catch this originally and it was due to the whole card after https://github.com/nunocoracao/blowfish/pull/2252 now using `absolute inset-0` and therefore preventing normal hover and requiring group hover.
- The `article` shortcode is broken, that was due to incompatibility of https://github.com/nunocoracao/blowfish/pull/2252 and https://github.com/nunocoracao/blowfish/pull/2297

If you need a more in-depth explanation, please let me know.

servedsmart:
@nunocoracao Conflcts are now resolved and were tested briefly. I also rebuilt `compiled/main.css`, but it might be wise to rebuild it once again before the release.

EDIT: I tested all files in `article-link/` for functionality on `exampleSite/`. Seems to work just like it did before the conflicts were introduced.


## Video showcasing the visual changes

At the very end, the somewhat buggy seeming behavior of main is also shown.
If you look at the bottom-left corner you can also see what the browser gives as a hint for what the href points to. Main doesn't use https://github.com/nunocoracao/blowfish/pull/2252 , therefore this is also different.

https://github.com/user-attachments/assets/8053b37b-7d27-4d92-8a4b-e5c256eeaca8

Issue #2184, opened 2025-05-29
Add matrix social link and replace mastodon with fediverse logo


louis-vgn:
**Is your feature request related to a problem? Please describe.**

Linking to matrix account is not supported.

For mastodon account, it should rather be the fediverse logo so that anyone from the fediverse (mastodon, pleroma, diapora, peertube, pixelfed, etc.) can link their account. It prevents adding every logo for anything related to fediverse.

**Describe the solution you'd like**
Implement matrix and fediverse social links.

**Additional context**
I can make the PR, it seems easy enough. I just need to have a pointer as to where to look at. I’ve never used tailwindcss.

insidemordecai:
I think the best option would be adding a new icon for the fediverse instead of replacing the existing mastodon icon this way one can pick what they want (either the individual federated platform icons or the fediverse icon). 

And you are right, it's easy enough and actually doesn't require knowledge of TailwindCSS. For where to look at, check the [Icons](https://blowfish.page/samples/icons/) page in the docs but the TLDR is you simply add the SVG file to `assets/icons/` and update the relevant docs. You can have a look at a similar PR I made - #1608

JoshuaACasey:
> For mastodon account, it should rather be the fediverse logo so that anyone from the fediverse (mastodon, pleroma, diapora, peertube, pixelfed, etc.) can link their account. It prevents adding every logo for anything related to fediverse.

I was about to come here to suggest adding Matrix. And Pixelfed. (Per your suggestion, I think it would be better to just have the individual icons for each app/service rather than one blanket icon for every fediverse app).

I'd also like to suggest adding: Keyoxide and Signal

insidemordecai:
@JoshuaACasey the Pixelfed icon is present already.

getting the Fediverse and Keyoxide icon to play nice with the themes and other icons took longer than i expected but i have managed and i'll create a pull request later. i believe they should be in the next available release 🤞

edit: oh, the Matrix icon will be included too

JoshuaACasey:
> [@JoshuaACasey](https://github.com/JoshuaACasey) the Pixelfed icon is present already.
> 
> getting the Fediverse and Keyoxide icon to play nice with the themes and other icons took longer than i expected but i have managed and i'll create a pull request later. i believe they should be in the next available release 🤞
> 
> edit: oh, the Matrix icon will be included too

[...]

PR #2354, opened 2025-07-18
✨ add matrix, fediverse, keyoxide and signal icons 

insidemordecai:
this adds icons requested in #2184 

this is how they would appear on the home page: 

Screenshot 2025-07-19 at 00 46 24

and this is how they would appear inline: 

Screenshot 2025-07-19 at 00 57 54


danielwagn3r:
Is it possible to add signal as supported social share icon a well?

insidemordecai:
@danielwagn3r that should be possible, you can open an issue and i'm sure someone will pick it up

PR #2412, opened 2025-08-09
fix bug in toc for articles

nunocoracao:


hudsonm62:
Hi @nunocoracao 

FYI This broke hiding the toc for every page that uses `showTableOfContents: false` on my blog.
What is the new setting for this?

PR #2433, opened 2025-08-19
:bug: Fix(rss.xml): Add missing `xml` frontmatter logic

racehd:
Frontmatter docs specify an `xml` parameter that was never implemented in the RSS template logic. Git history shows this was documented since creation but never coded (only in sitemap.xml but not rss.xml)

- Add filter logic to exclude pages when `xml: false` is explicitly set. Use explicit loop instead of one-liner `where` filter to ensure only pages with `xml: false` are excluded (not pages with missing or other values).

- Update docs to reference "RSS feed" for better searchability.

Technical note: Hugo's `build.list: never` removes pages from all collections including RSS, but this `xml` parameter allows excluding from RSS while keeping in other page lists.


racehd:
Looking at it further, i do see the `xml` param used in sitemap.xml, not `rss.xml` will close this PR and open discussion #2435

Issue #2448, opened 2025-08-31
[Bug]: Build failed when using an external link for featureimage in the front matter


eallion:
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Describe the Bug

Build failed when using an external link for featureimage in the front matter.

```
error calling GetRemote: failed to resolve media type for remote resource 
```

And not a networking issue.

### To Reproduce

1. Front Matter:

```
---
...
featureimage: https://images.xxx.com/directus/files/629e8e6c-9342-4948-9d63-b73590c88e08.jpg
---
```

<img width="980" height="398" alt="Image" src="https://github.com/user-attachments/assets/40b49305-71af-48ef-8804-3027abf1f21c" />

2. Build Hugo 

```
hugo version
hugo v0.148.2-40c3d8233d4b123eff74725e5766fc6272f0a84d+extended windows/amd64 BuildDate=2025-07-27T12:43:24Z VendorInfo=gohugoio
``` 
```
hugo --gc --minify
```
```
ERROR render of "/blog" failed: "D:\github.com\xxx.com\themes\blowfish\layouts\_default\list.html:77:17": execute of template failed: template: list.html:77:17: executing "main" at <partial "article-link/simple.html" .>: error calling partial: "D:\github.com\xxx.com\themes\blowfish\layouts\partials\article-link\simple.html:46:30": execute of template failed: template: _partials/article-link/simple.html:46:30: executing "_partials/article-link/simple.html" at <resources.GetRemote>: error calling GetRemote: failed to resolve media type for remote resource "https://images.xxx.com/directus/files/629e8e6c-9342-4948-9d63-b73590c88e08.jpg"
[...]

Issue #1938, opened 2025-01-25
Frontmatter: `featureimage` internal url


bitSheriff:
**Is your feature request related to a problem? Please describe.**
My problem is that the frontmatter `featureimage` only allows *external urls*. I want to set a feature image for more than one post without duplicating the `.png` file into every folder.

**Describe the solution you'd like**
it would be very nice to support internal ulrs like `/assets/img/feature.png` too.

**Describe alternatives you've considered**
I considered putting these feature images into the static directory and setting the external url to my page. BUT therefore it cannot be ported easily to another domain because the domain is hard coded in these urls.

Thank you very much

BasilSkrnk:
I’m interested in this as well.

This solution, I’ve come up to this too.
> I considered putting these feature images into the static directory and setting the external url to my page. BUT therefore it cannot be ported easily to another domain because the domain is hard coded in these urls.
Yet, for me, that does not always work. E.g. I have a couple of projects that live mostly in the local network, before the ‘stable’ version is pushed to the internet. When I add a new article with the featured image, the external link would start working only when it’s deployed world-wide. While it lives in the intranet (where we test things), it won’t appear till the article appears on the public version of the project. 

Obviously, it won’t work while developing, too.

realcatgirly:
I'm also interested in this since I reuse a lot of feature images as placeholders on multiple articles

eallion:
Same

PR #2451, opened 2025-08-31
✨ Feat(article-link, hero): improve image handling

ZhenShuo2021:
Scope: article-link, hero

Summary:
Refactor and improve rendering logic for article-link and hero modules, including template separation, conditional image handling, and priority management.

Key changes:

1. Remove unused `.Params.images` and `.Site.Params.images`
2. Skip fetching images when `hideFeatureImage` is true
3. Explicitly set featured image filename priority
4. Give the `featureimage` frontmatter the highest priority to prevent unintended overrides (it was confusing for users who set frontmatter but still had it overridden by the filename)
5. Support internal links for `featureimage`
   1. Closes #2448
   2. Marks #1938 as completed, not as not planned
6. 7498f45 Using [::before](https://kittygiraudel.com/2022/04/02/accessible-cards/) to implement article-link hover.
   1. Benefits are as described in the commit.
   2. Additionally, it enables the a11y underline feature to add title underlines without modifying the JS file.
   3. Marks #1729 as completed, not as not planned
7. 6768f8e Use semantic HTML

PR #2452, opened 2025-08-31
feat: external image for front matter featureimage

eallion:
I'm not sure if this is the most logical approach, this might also bring new issues, but it fixed my issue. 

Now, if I provide a `featuredimage` URL in the Front Matter and set `disableFeaturedImageProcessing` to `true` in `params.toml`, the featured image in both cards and lists will display the URL directly.

IMO: These images have already been optimized and deployed to a CDN image-bed.

Solved #2448 

Feat for #1938


ZhenShuo2021:
I think `disableFeaturedImageProcessing` can be achieved with the following settings:

```
disableImageOptimization = true
disableImageOptimizationMD = false
```

Also, handling internal links for `featureimage` can be addressed via #2451. This setup should work for most use cases, but it may be helpful to verify it in your environment.

eallion:
Closed. Not good.

Issue #1808, opened 2024-10-22
Timeline support markdown


Weaxs:
**Is your feature request related to a problem? Please describe.**
Timeline shortcode not support markdown now
![image](https://github.com/user-attachments/assets/2b86fef4-3b4f-4861-ae03-509fcbd8a1a1)

**Describe the solution you'd like**
-

**Describe alternatives you've considered**
-

**Additional context**
-

wermos:
Can you share a MWE (minimal working example)?

It _might_ have something to do with a PR I made (#1515).

Weaxs:
> Can you share a MWE (minimal working example)?
> 
> It _might_ have something to do with a PR I made ([#1515](https://github.com/nunocoracao/blowfish/pull/1515)).
> 
> But it still seems to work normally on the [Blowfish docs page](https://blowfish.page/docs/shortcodes/#timeline).
> 
> Can you share what version of Blowfish and Hugo you were using?

like this:
```
{{< timeline >}}

{{< timelineItem icon="star" header="2023.07" badge="Jina Embeddings" >}}
[[2023.07] Jina Embeddings: A Novel Set of High-Performance Sentence Embedding Models](https://arxiv.org/abs/2307.11224):<br/> The first version of Jina's Embedding is based on the **T5 model ** ([*Exploring the Limits of Transfer Learning with a Unified Text-to-Text Transformer. [2019.10]*](https://arxiv.org/abs/1910.10683)) for pre-training, and secondly, the Jina team designed a new dataset specifically for training the Embedding model.
{{< /timelineItem >}}

{{< timelineItem icon="star" header="2023.10" badge="Jina Embeddings 2" >}}
[[2023.10] Jina Embeddings 2: 8192-Token General-Purpose Text Embeddings for Long Documents](https://arxiv.org/abs/2310.19923):<br/> Jina's second version of Embedding is pre-trained on the basis of the    pretrained on the **Bert model**, using linear bias Attention ([**ALiBi**](https://arxiv.org/abs/2108.12409)) to increase the contextual information of the model input; secondly, the text data is used to fine-tune the model's embedding ability; finally, **adversarial example fine-tuning** is added to further improve the model's ability.
{{< /timelineItem >}}

{{< timelineItem icon="star" header="2024.09" badge="jina-embeddings-v3" >}}
[...]

PR #2465, opened 2025-09-07
✨ Feat(timeline): support markdown content

ZhenShuo2021:
Fix this issue because it is a quite common need.

Should mark #1808 as completed, not as not planned.


pfig (approved):
Terrific, thank you.

PR #2488, opened 2025-09-17
Update users.json

HeTongRe4per:
Update users.json
Update my website that was mistakenly deleted.


nunocoracao:
This PR is replacing an existing user. Will close. 
If the website was removed was because it was down during one of the release cycles. Please do add it again to the end of the list if you want to.

PR #2513, opened 2025-09-29
fix bug: (🐛) adding me instead of mr and compiling tailwind

abo3skr2019:
this PR is based on #2393 where they used logical properties instead of ltr and rtl i just extended that to simple.html which had an issue where there was no margin between the image and the link partial

fix:
- simple.html article image padding

resolves #2393


nunocoracao:
Closing this PR as it included a user.json change and it was targeting main - please target dev branch

AFAskar:
Roger that

PR #2534, opened 2025-10-11
✨ Feat: remove jquery

ZhenShuo2021:
Replaced jQuery methods with native JavaScript API.

Putting this up as a draft PR for now to see if anyone has feedback or to check if anything was missed.

## Changes
- Remove the jquery library (vendor.html, package.json)
- The `highlightCurrentMenuArea` feature (header/basic.html)
- The `gallery` shortcode (assets/js/shortcodes/gallery.js)


nunocoracao:
This is great if you can pull it off. I suspect some hidden corners will break. We should test this one well.

ZhenShuo2021:
I've tested this by observing the debug console to see if it throws any errors that indicate jQuery dependencies. Some optional features like `smartTOCHideUnfocusedChildren` were already rewritten to vanilla JS in my earlier PR #2202.

ZhenShuo2021:
I think we should keep this here a bit longer since we've had many changes in recent releases. It would help ensure a more stable experience for users.

PR #2546, opened 2025-10-14
🐛Fix(rtl): Image Padding Issues V2

AFAskar:
this PR is based on #2393  where they used logical properties instead of ltr and rtl i just extended that to simple.html which had an issue where there was no margin between the image and the link partial

fix:
- simple.html article image padding

Couldn't compile tailwind due to issues with the ltr / rtl directives and the removal of the cli from the package json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hugo-blowfish-theme",
-  "version": "2.91.0",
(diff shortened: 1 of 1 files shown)


AFAskar:
Fixing conflict

PR #2582, opened 2025-10-28
docs: remove duplicate list.showTableOfContents

kv-chiu:
image

image


nunocoracao:
These drive different behaviors it's not a duplication

kv-chiu:
> These drive different behaviors it's not a duplication

Sorry for the delayed response—I just saw your message from November 4th. It looks like there might have been a misunderstanding about my PR, and I'd like to clarify:

I didn't mean that `list.showTableOfContents` and `article.showTableOfContents` are duplicates in the documentation. Rather, `list.showTableOfContents` appears twice in the docs, and the second instance has the exact same description as `article.showTableOfContents`—which seems quite puzzling.

I've tested both parameters, and they correctly control `enableToc` for `list` and `single` pages respectively. So the duplicate entry for `list.showTableOfContents` in the documentation isn't an intentional feature, is it?

Issue #2690, opened 2025-12-31
[Bug]: Reference and footnote style links are not rendered in alerts


campfred:
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Hugo Version Check

- [x] My Hugo version is within the supported range

### Describe the Bug

Upon learning to write articles using this theme, I found that [reference-style links][markdownguide-basicsyntax-referencestylelinks] and [footnotes][markdownguide-extendedsyntax-footnotes] don't work inside `alert` shortcode blocks.

They simply appear as their source code (`[text][reference]` and `text[^#]` respectively).

[markdownguide-basicsyntax-referencestylelinks]: https://www.markdownguide.org/basic-syntax/#reference-style-links "Markdown Guide: Basic Syntax > Reference-style links"
[markdownguide-extendedsyntax-footnotes]: https://www.markdownguide.org/extended-syntax/#footnotes "Markdown Guide: Extended Syntax > Footnotes"

### To Reproduce

1. Put this example code snippet in a page
   > ```markdown
   > {{<alert>}}
   > This is a test alert with a [reference-style link][markdownguide-basicsyntax-referencestylelinks] and a footnote[^8].
   > {{</alert>}}
   > 
   > [markdownguide-basicsyntax-referencestylelinks]: https://www.markdownguide.org/   > basic-syntax/#reference-style-links "Markdown Guide: Basic Syntax > Reference-style links"
   > [^8]: https://www.markdownguide.org/extended-syntax/#footnotes "Markdown Guide: Extended Syntax > Footnotes"
2. Run Hugo server
3. Open the page
4. Observe the links being simply written as their source code

### Expected Behaviour

Alert's content gets rendered with the requested links.

### Screenshots

<img width="526" height="123" alt="Image" src="https://github.com/user-attachments/assets/a029b610-acbb-47cf-96c2-ad6c5ffff458" />

#### Test setup 1

- OS: `Bazzite Linux 43`
- Browser: `Zen Browser`
- Version: `1.17.15b (Firefox 146.0.1)`

#### Test setup 2

- OS: `Windows 11`
- Browser: `Helium Browser`
- Version: `0.7.7.1 (Chromium 143.0.7499.169)`

### Hugo Version

```console
hugo v0.152.2+extended+withdeploy linux/amd64 BuildDate=2025-10-24T15:31:49Z VendorInfo=brew
```

### Blowfish Version

```console
v2.94.0?
[...]

PR #2691, opened 2025-12-31
✨ Feat(admonition): support custom icon

ZhenShuo2021:
This PR essentially makes the admonition a full replacement for the alert shortcode, as both now support arbitrary icons.

I'm considering adding a deprecation tag to the alert shortcode. The admonition (render hook) is significantly more powerful than the shortcode and has the following advantages:

1. Markdown syntax is portable, while shortcodes are specific to Hugo.
2. General users find it difficult to choose a good-looking color for the alert shortcode. The colors of the admonition are curated for both light and dark modes, whereas the alert shortcode does not support light/dark mode settings.
3. The colors of the admonition can be customized via [CSS variables](https://github.com/nunocoracao/blowfish/blob/v2.94.0/assets/css/components/admonition.css).
4. Footnotes inside shortcodes are not supported (#2690).
5. The alert shortcode does not render well with multi-paragraph content.
    image

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -94,6 +94,9 @@ The alert sign (`+` or `-`) is optional to control whether the admonition is fol
 > **GitHub types:** `NOTE`, `TIP`, `IMPORTANT`, `WARNING`, `CAUTION`  
 > **Obsidian types:** `note`, `abstract`, `info`, `todo`, `tip`, `success`, `question`, `warning`, `failure`, `danger`, `bug`, `example`, `quote`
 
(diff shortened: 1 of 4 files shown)


nicholas-alonzo:
@ZhenShuo2021 , I was thinking of adding a feature request for this so good to know this was considered. On a similar note I tried to include an icon with the label of the the new Tabs shortcode, but it wouldn't render it. Thoughts on the ability to include an icon there?

nunocoracao:
lgtm - let's leave the alert shortcodes for now they are easy to maintain 
If we plan to remove them or simplify our lifes I would like to make that change backwards compatible with someone using the existing shortcodes

PR #2694, opened 2025-12-31
🐛 Fix(rss): lastBuildDate and description

### LINKED ISSUE #2693: RSS feed generates invalid XML due to safeHTML preventing XML escaping
## Description

The RSS template uses `| safeHTML` which prevents Hugo's default XML escaping, producing invalid XML when article titles or descriptions contain special characters.

## Location

`layouts/_default/rss.xml` lines 38 and 43

## Current code

```go
<title>{{ .Title }}</title>
...
<description>{{ .Summary | plainify | htmlUnescape | safeHTML }}</description>
```

## Problem

The `safeHTML` filter tells Hugo not to escape the output for the target format. This means:
1. `htmlUnescape` correctly converts HTML entities (e.g., `&rsquo;` → `'`)
2. But ampersands in content (e.g., "Bubble Tea & Mochi") are NOT escaped to `&amp;`
3. Result: Invalid XML like `<title>East & West</title>`

## Error

```
xmlParseEntityRef: no name
```

## Suggested fix

Remove `safeHTML` - keep `htmlUnescape` to convert HTML entities, but let Hugo's default XML escaping handle special characters:

```go
<title>{{ .Title | plainify | htmlUnescape }}</title>
...
<description>{{ .Summary | plainify | htmlUnescape }}</description>
```

This allows Hugo to properly escape `&` → `&amp;`, `<` → `&lt;`, etc.

## Version

v2.94.0

ZhenShuo2021:
Closes #2693

Align usage with Hugo

https://github.com/gohugoio/hugo/blob/v0.153.4/tpl/tplimpl/embedded/templates/rss.xml

### DIFF
--- a/layouts/_default/rss.xml
+++ b/layouts/_default/rss.xml
@@ -23,7 +23,7 @@
     {{ if .Site.Params.footer.showCopyright | default true -}}
     <copyright>{{ with replace .Site.Params.copyright "{ year }" now.Year }}{{.}}{{ else }}© {{ now.Format "2006" }} {{ .Site.Params.Author.name }}{{- end }}</copyright>
     {{- end }}
-    {{ if not .Date.IsZero }}<lastBuildDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</lastBuildDate>{{ end }}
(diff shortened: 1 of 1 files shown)


ilmax:
I wonder if it's worth starting to add some unit test coverage to make it difficult to (re)introduce this kind of bugs. I seem to remember that there was a poll about it a while ago but couldn't find the issue anymore.

ZhenShuo2021:
The problem is that Hugo behaves like a dynamic language, so in many cases the only way to know whether the output is correct is to render it and inspect it manually.

Adding a test for RSS is easy, but that's much harder to do for regular templates. With our current resources, my concern is ending up with tests that don't actually catch the issue.

PR #2760, opened 2026-01-29
🐛 Fix: replace local typeit with CDN

ZhenShuo2021:
TypeIt is licensed under GPLv3. Distributing it as a vendor copy constitutes redistribution, which forces Blowfish to also adopt the GPLv3 license. By switching to a CDN link, we avoid this redistribution issue.

One drawback of this change is that users in China may be unable to use this feature, as most CDN providers are blocked by the GFW.


nunocoracao:
THis branch includes some changes from the other PR
Let's not change type it for now it if blocks people from using it
I'll think about the licensing topic

PR #2773, opened 2026-02-06
feat: Add captions argument to carousel shortcode

alxhslm:
I wanted to be able to add captions to a carousel of images, but realised I couldn't. 

Thought about changing the API to support using the `figure` shortcode in the body (like in #2771), but this would've been a breaking change. 

Instead, I've chose the simple option to add a new optional `captions` argument.


ZhenShuo2021:
I tested this change and noticed it causes horizontal scrolling. Could you take a look?

alxhslm:
> I tested this change and noticed it causes horizontal scrolling. Could you take a look?

Good spot! Fixed now

PR #2796, opened 2026-02-17
feat(comments): add new CSS for nested comments, avatars, and metadata

manio:
Hi there,
Just to be upfront, I feel much more comfortable with system/backend stuff than frontend, so I hope what I’ve put together here isn’t a total disaster :)

In short: I used a neat WordPress-to-Hugo importer ([wp2hugo](https://github.com/ashishb/wp2hugo)) that can import comments from WordPress. Unfortunately, it doesn’t include any styles, so you have to handle that yourself. This PR takes care of that.

Here’s what it looks like before:
image

…and after applying this PR:
image

I hope you also like it like me :)
Colors adapts to the color scheme configured in blowfish.

I’d really appreciate it if this could be included in your beautiful theme :)
That way others can benefit from it too.

### DIFF
--- a/assets/css/components/comments.css
+++ b/assets/css/components/comments.css
@@ -1,0 +1,56 @@
+#comments,
+#comments ul {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+#comments li {
+  position: relative;
+  padding-left: 1rem;
+}
+
+.children-comments {
+  margin-left: 2rem;
+  padding-left: 1.5rem;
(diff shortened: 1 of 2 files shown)


nunocoracao:
Sorry @manio this does't seem to be reusable. 
This is just styling for some classes you created in your instance of the theme.

manio:
I believe it was not for "my instance" but for the ([wp2hugo](https://github.com/ashishb/wp2hugo)) project's output, I described above, but OK.

PR #2841, opened 2026-03-15
Fix Umami event name for long page titles

### LINKED ISSUE #2840: [Bug]: Umami track event can exceed Umami's event name limit when `og:title` is long
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Hugo Version Check

- [x] My Hugo version is within the supported range

### Describe the Bug

First of all, thanks for the built-in Umami support in Blowfish. While testing it, I ran into a small compatibility issue with `enableTrackEvent`.

When `umamiAnalytics.enableTrackEvent` is enabled, Blowfish injects a script that sends a custom Umami event after the tracker script has loaded.

The event name is currently built from the Open Graph type and the full Open Graph title:

```html
{{ if .Site.Params.umamiAnalytics.enableTrackEvent | default true }}
  <script type="text/javascript">
    document.querySelector('script[data-id="umami-script"]').addEventListener("load", function () {
      const type = document.head.querySelector('meta[property = "og:type"]').getAttribute("content");
      let title = document.head.querySelector('meta[property = "og:title"]').getAttribute("content");
      let url = document.head.querySelector('meta[property = "og:url"]').getAttribute("content");
      umami.track(type + ":" + title, { url: url });
    });
  </script>
{{ end }}
```

This works for shorter titles, but it fails when `type + ":" + title` exceeds Umami's maximum event name length.

In that case, normal pageviews still appear to be tracked, but the additional custom event request returns HTTP 400 in the browser console, so the Events counter in Umami stays at 0.

Example failing event name:

```text
article:Clericó: der fruchtige Klassiker für Feste und Sommerabende
```

Example response from Umami:

```json
{
  "error": {
    "message": "Bad request",
    "code": "bad-request",
    "status": 400,
    "errors": [],
    "properties": {
      "payload": {
        "errors": [],
        "properties": {
          "name": {
            "errors": [
              "Too big: expected string to have <=50 characters"
            ]
          }
        }
      }
    }
  }
}
```

### To Reproduce

1. Configure Umami in Blowfish and enable track events:

```toml
[umamiAnalytics]
  websiteid = "MY-WEBSITE-ID"
  domain = "cloud.umami.is"
  dataDomains = "MY-DOMAIN.COM"
  scriptName = "script.js"
  doNotTrack = true
  enableTrackEvent = true
```

2. Create or open any page whose `og:title` is long enough to push the generated event name above Umami's limit.
3. Load that page in the browser.
4. Open the browser dev tools and inspect the network request to Umami.
5. Observe the POST request to the Umami API returning HTTP 400.

### Expected Behaviour

Blowfish should avoid generating Umami event names that can become invalid for longer page titles.

A short, stable event name would probably be more robust here, for example:

```js
umami.track(type + ":view", { title: title, url: url });
```

or another fixed event name with the title stored in the payload instead of the event name itself.

This would keep the event valid, avoid console errors, and also make the event grouping more stable over time.

### Screenshots

No screenshots attached. The browser console/network panel shows a failed POST request to Umami with HTTP 400.

### Platform

- OS: Linux
- Browser: Firefox
- Version: 148.0

### Hugo Version

```console
hugo v0.155.3+extended+withdeploy linux/amd64 BuildDate=unknown
```

### Blowfish Version

```console
v2.100.0
```

### Additional Context

From the outside this looks less like an Umami configuration issue and more like a small edge case in the Blowfish Umami partial.

The request is sent correctly, but the generated custom event name can become invalid when the page title is long. A fixed technical event name plus contextual data in the payload seems like the cleanest fix.

If this approach makes sense, I would be happy to follow up with a PR.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

aplgr:
## Summary

This PR changes the Umami tracking event name from a dynamic `type:title` value to a stable `type:view` value and moves the page title into the event data payload.

## Problem

The current implementation uses the full `og:title` as part of the Umami event name:

```js
umami.track(type + ":" + title, { url: url });
```

Umami limits event names to 50 characters. For pages with longer titles, the request fails with HTTP 400 and the event is rejected.

## Fix

This PR changes the event call to:

```js
umami.track(type + ":view", { title: title, url: url });
```

This keeps the event name short and stable while still preserving the page title and URL in the event data.

## Result

- avoids HTTP 400 responses for long titles
- keeps Umami event names within the documented limit
- preserves useful metadata in the payload

## Related issue

Closes #2840

### DIFF
--- a/exampleSite/content/users/users.json
+++ b/exampleSite/content/users/users.json
@@ -1036,4 +1036,12 @@
       "Blog"
     ]
   }
+  {
+    "title": "Robin Fairchild",
+    "url": "https://fairchild26.github.io/",
+    "tags": [
+      "Personal site",
+      "Blog"
+    ]
+  }
 ]
--- a/layouts/partials/analytics/umami.html
+++ b/layouts/partials/analytics/umami.html
@@ -23,7 +23,7 @@
       const type = document.head.querySelector('meta[property = "og:type"]').getAttribute("content");
       let title = document.head.querySelector('meta[property = "og:title"]').getAttribute("content");
       let url = document.head.querySelector('meta[property = "og:url"]').getAttribute("content");
-      umami.track(type + ":" + title, { url: url });
+      umami.track(type + ":view", { title: title, url: url });
     });
   </script>
 {{ end }}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hugo-blowfish-theme",
-  "version": "2.100.0",
+  "version": "2.101.0",
   "description": "Blowfish theme for Hugo.",
   "scripts": {
     "postinstall": "vendor-copy",


nunocoracao:
Thanks for flagging this! The Umami event name fix is a good catch and I'll be applying it directly. However, I'm closing this PR because it bundles unrelated changes (version bump and user addition to users.json) alongside the bug fix. In the future, please keep PRs focused on a single change — it makes review much easier. Thanks for contributing!

PR #2848, opened 2026-03-19
added events

Cody935:


Cody935:
Working

nunocoracao:
Thanks for the contribution! However, this looks like custom CSS for a specific site layout rather than a general-purpose theme feature. Since it targets a very specific .events-container pattern without an accompanying shortcode, configuration, or documentation, it's better suited as a site-level customization (e.g., in your own custom CSS or extend-head partial) rather than something bundled into the theme. Feel free to reopen if you'd like to propose it as a more general shortcode/feature with docs!

Issue #2850, opened 2026-03-22
[Bug]: HTML in the `title` not stripped before insertion to `head`


nk9:
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Hugo Version Check

- [x] My Hugo version is within the supported range

### Describe the Bug

It's perfectly valid to include HTML in the title of your content, and it will be rendered in the page:

```yaml
---
title: "Why <u>you</u> should be using Blowfish"
---
```
<img width="606" height="80" alt="Image" src="https://github.com/user-attachments/assets/24ac9e7b-6e5d-429a-9e01-ebfbe6c6c042" />

However, this code isn't removed in the page title by Blowfish:

<img width="223" height="43" alt="Image" src="https://github.com/user-attachments/assets/91b5e247-2870-4769-8af3-18f8afc598ae" />

It should also be removed from the `meta` title as well. This just requires putting `plainify` before `emojify`.

### To Reproduce

Put HTML in your title

### Expected Behaviour

Plain text <code><title></code>

### Platform

- OS: macOS 15
- Browser: Firefox
- Version: N/A

### Hugo Version

```console
hugo v0.158.0+extended+withdeploy darwin/arm64 BuildDate=2026-03-16T17:42:04Z VendorInfo=Homebrew
```

### Blowfish Version

```console
2.100.0
Would be nice to have an indication of where to get this number from in the bug creation flow!
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nunocoracao:
Fixed in PR #2851 — `.Title` is now piped through `plainify` before `emojify` in the `<head>`, which strips any HTML before it reaches the page title.

PR #2851, opened 2026-03-22
[#2850] Run plainify on the title value before emojify

nk9:
Address #2850


nk9:
Question: should this same sanitization be applied to the site title as well?

PR #2852, opened 2026-03-23
:sparkles: Implemented category badges in secondary color and better granularity for showing categories or tags

MaikelChan:
Hi, it's been a while since my last PR!

Previously, there was an option to show only category badges with `showCategoryOnly`. This allowed to either show both categories or tags, or only categories. But if you wanted to show only tags, that wasn't an option.

I implemented new settings that allow just that: `showCategories` and `showTags`. Both are enabled by default but depend on the already existing `showTaxonomies` as a global setting for both types of taxonomies. So this new setup is backwards compatible and shouldn't break the appearance of existing sites.

I also added the `showCategoriesInSecondaryColor` setting, which allows to show the category badges in a secondary color of the template. This allows the user to better distinguish between both types of badges, specially on a site that uses both extensively. It's set to false by default also to preserve backwards compatibility.

Here's an image of how it looks:

Badges

I implemented this by creating a new `badge-secondary` partial, but initially wanted to pass a variable of whether the taxonomy is a tag or a category to the existing `badge` partial, but I wasn't able to get it to work.

Let me know if that way is preferable and if you have a recommended way on how to pass that variable to the partial.


MaikelChan:
I just removed the badge-secondary partial and now I pass the variable that defines whether to show the badge with a primary color or a secondary color with a dictionary.

PR #2871, opened 2026-03-27
Add minimal cookie notice MVP

aplgr:
This PR adds a minimal built-in cookie notice banner for Blowfish.

Scope:
* optional cookie notice banner
* simple dismiss flow
* configurable browser storage (`localStorage` or `sessionStorage`)
* optional image support
* multilingual privacy policy link via `privacyPolicyPageRef`
* external JS/CSS assets
* docs and example config updates

This is intentionally an MVP and not a full consent manager or compliance framework. The goal is to provide a small built-in base that can be extended later if needed.


nunocoracao:
Thanks for the thorough work on this! The implementation looks well-scoped as an MVP. However, this PR is too large to review and merge as-is — 52 files including all i18n strings, new JS/CSS assets, config, and docs. Could you break this into smaller, incremental PRs? For example: 1) core banner component + config, 2) i18n strings, 3) docs/example site updates. That would make it much easier to review and merge. Happy to revisit once it's split up!

PR #2891, opened 2026-04-07
Migrate to Hugo v0.146.0+ new template system

### LINKED ISSUE #2785: [Feature|FutureBug]: Migrate to Hugo's "new template system" (v0.146.0+)
Blowfish theme as of now ([v2.98.0](https://github.com/nunocoracao/blowfish/releases/tag/v2.98.0)) has not yet migrated to Hugo's [new template system](https://gohugo.io/templates/new-templatesystem-overview/). Supporting that shouldn't be too difficult.

Backward compatibility might probably become future issues, otherwise.

EricGusmao:
Fixes #2785

Migrates Blowfish to Hugo's new template system introduced in v0.146.0. See the [Hugo new template system overview](https://gohugo.io/templates/new-templatesystem-overview/) for details.

## Summary

- Restructures `layouts/` to comply with Hugo v0.146.0+'s new template system: removes `_default/`, renames `partials/` → `_partials/`, `shortcodes/` → `_shortcodes/`, and promotes render hooks to `_markup/`
- Renames `index.html` → `home.html`, `terms.html` → `taxonomy.html`, and `index.json` → `home.json` to match new Hugo page-kind template naming
- Replaces deprecated `{{ template "_internal/opengraph.html" }}` / `{{ template "_internal/twitter_cards.html" }}` calls with `{{ partial "..." }}`, and updates all `templates.Exists` / `printf` paths from `"partials/..."` to `"_partials/..."`
- Applies the same `_partials/` / `_shortcodes/` renames to `exampleSite/layouts/`
- Bumps minimum Hugo version in `config.toml` to `0.146.0`


nunocoracao:
Thanks for working on this! Unfortunately, this PR introduces breaking changes for existing users — anyone who overrides partials (e.g. `layouts/partials/extend-footer.html`, `layouts/partials/extend-head.html`, `layouts/partials/favicons.html`, etc.) would have their customizations silently stop working after upgrading. We don't accept breaking changes unless they're strictly required to keep supporting the project (e.g. a dependency deprecating a feature we rely on). Hugo's new template system with `_partials` is supported but the old `partials` path still works, so this migration isn't required yet. If Hugo ever removes support for the old paths, we'd revisit this. Thanks for understanding!

PR #2935, opened 2026-05-10
✨ Add opt-in client-side language redirect

aplgr:
## Summary

This PR adds an optional client-side language redirect feature for multilingual Blowfish sites.

Related discussion: https://github.com/nunocoracao/blowfish/discussions/2934

The feature is fully opt-in and disabled by default. Existing sites keep their current behavior unless they explicitly enable the new `languageRedirect` configuration.

## Problem

Blowfish already provides a language dropdown for multilingual sites, but visitors still land on the configured default content language first.

For static multilingual sites without server-side language negotiation, this means that a visitor whose browser prefers another supported language has to switch manually, even when a translated version already exists.

## Solution

This PR introduces a small client-side redirect helper that:

- detects the browser language on first visit
- redirects only when a matching translation exists
- stores explicit language selections from the existing language dropdown in `localStorage`
- respects the stored language preference on later visits
- avoids redirect loops
- does not require cookies or any server-side component

The browser-language redirect is limited to home pages by default to avoid surprising users who open deep links. Stored language preferences can still be respected on translated pages when a matching translation exists.

The matching logic preserves locale subtags where available, so sites using variants such as `pt-BR` / `pt-PT` or `zh-Hans` / `zh-Hant` can still honor the visitor’s exact browser preference before falling back to a base language match.

## Configuration

The feature can be enabled with:

```toml
[languageRedirect]
  enabled = true
  storageKey = "blowfish_preferred_language"
  fallbackLanguage = "en"
  browserRedirectHomeOnly = true
  storedLanguageRedirect = true
````

## Backward compatibility

This is disabled by default.

No existing multilingual behavior changes unless the site owner explicitly enables:

```toml
[languageRedirect]
  enabled = true
```

## Notes

The implementation is intentionally client-side only and uses the existing Hugo/Blowfish multilingual data, especially `.AllTranslations`, instead of introducing any server-side requirements or external dependencies.


PierreMesure:
Great addition @aplgr!

I just tried locally on my website and I can't avoid the redirect when on a specific URL. This is my config:

```toml
[languageRedirect]
  enabled = true
  browserRedirectHomeOnly = true
```

My browser language is `en` and my website is also showing in English. If I try to load `http://localhost:1313/sv/projects/politix/`, the `/sv` gets automatically removed and I end up on the English version of the page.

EDIT: I added `storedLanguageRedirect = true` and it now works as expected! 😊

EDIT 2: I just noticed that with the 3 parameters mentioned above, it's impossible to manually change the language from the homepage. Let's say my browser language is set to Swedish and I try to change to English in the language switcher, it will automatically reload the Swedish page since it doesn't store any language preference in localStorage.

I think the most intuitive behaviour for me would be:
- Redirect to browser language when first visiting the homepage
- Allow the user to change the language manually from any page, including the homepage
- No redirect from deep links

PR #2975, opened 2026-06-08
✨ Add basic microformats2 support

tb1337:
Adds opt-in [[microformats2](https://microformats.org/wiki/microformats2)](https://microformats.org/wiki/microformats2) markup behind a new `enableMicroformats2` param (default `false`).

Everything is gated, so default output is unchanged and therefore non-breaking.

**Changes:**

- `single.html`: post as `h-entry` (`p-name`, `e-content`, `dt-published`/`dt-updated`, `u-url`/`u-uid`, `p-author h-card`)
- `list.html`: article-grid sections get `h-feed`
- `article-link/card.html` + `simple.html`: items as `h-entry` (`u-url`, `p-name`, `dt-published`, `p-summary`)
- `footer.html`: hidden representative `h-card` from `params.author` (email omitted, matching the `rel="me"` handling in `head.html`)
- docs + exampleSite `params.toml`, cross-linked with `verification.fediverse`

**Decisions:**
I put this behind a dedicated `enableMicroformats2` switch instead of reusing `verification.fediverse`. Different scope (one `<head>` meta tag vs site-wide markup) and independent of each other (the `h-card` reads `params.author`, not the handle).

This is a first step. I'd like to contribute further syndication / IndieWeb features over time.

### DIFF
--- a/exampleSite/config/_default/params.toml
+++ b/exampleSite/config/_default/params.toml
@@ -13,6 +13,7 @@ enableA11y = false
 enableSearch = true
 enableCodeCopy = true
 enableStructuredBreadcrumbs = true
+enableMicroformats2 = false
 
 replyByEmail = false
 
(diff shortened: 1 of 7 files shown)


tb1337:
nvm, have to do some more work on this

PR #2980, opened 2026-06-17
Feat(mobile menu): Add a theme parameter to determine whether mobile nested menu items are expanded by default.

yozlog:
## Summary

Set the default appearance of the mobile menu as shown in the image below.

mobile-menu

## Changes

- Add a new parameter to `params.toml` to control whether nested menu items on mobile devices are expanded by default.
- Update the mobile menu behavior to respect the new configuration.
- Document the new parameter in the configuration docs.

## Notes

I'm not sure whether i should update the configuration docs or not, but I did anyway. 

Only the English and Simplified Chinese versions of the configuration docs have been reviewed. The other translations were generated using AI, and I cannot guarantee their grammatical or semantic accuracy.

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -47,6 +47,7 @@ forgejoDefaultServer = "https://v11.next.forgejo.org"
 
 [header]
   layout = "basic" # valid options: basic, fixed, fixed-fill, fixed-gradient, fixed-fill-blur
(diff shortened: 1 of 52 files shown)


nunocoracao:
Love the mobile changes - but the PR includes a bunch of other stuff - 52 files changed is too much. Please split into different PRs. I would really like to get the mobile change in first. Quick question, why is it not collapsed by default?

PR #2988, opened 2026-07-01
feat: add Knocket live chat widget (opt-in)

rrivory:
Adds Knocket as an optional live chat widget using the existing analytics provider pattern. Knocket is a 100% free live chat widget — free forever, no ads, no seat limits. Disabled by default; renders only when knocketChat.identifier is set. More info: https://trtc.io/solutions/knocket


nunocoracao:
Closing this one, adding 3rd party dependencies requires either more user traction or a discussion. If this is only for one user feel free to use the overrides available in Hugo to add this to your website on top of Blowfish.

PR #3001, opened 2026-07-10
fix: article shortcode build deadlock on mutually-referencing articles

### LINKED ISSUE #2791: [Bug]: The `{{< article >}}` shortcode doesn't handle mutual references
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Hugo Version Check

- [x] My Hugo version is within the supported range

### Describe the Bug

The site fails to build if two articles have mutually linking `article` shortcodes.

### To Reproduce

1. Create articles A and B
2. In article A, create a shortcode to article B
3. In article B, create a shortcode to article A
4. Try to build the site (live preview)
5. Hugo hangs and stop responding to requests

### Expected Behaviour

The expected behaviour would be to have the shortcode working in both articles.
Intuitively, this should be possible given that the "article series" feature already support mutual links.

### Screenshots

_No response_

### Platform

- OS: Fedora (GNU/Linux)
- Browser: Firefox
- Version: Latest for both

### Hugo Version

```console
hugo v0.152.2+extended linux/amd64 BuildDate=2025-12-14T00:00:00+00:00 VendorInfo=Fedora:0.152.2-1.fc43
```

### Blowfish Version

```console
**How do I get the version number!?**
It should however be v2.98.0 (I believe) as I updated it before trying to build the site via `blowfish-tools`.
```

### Additional Context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

muzimu:
Fixes #2791

## Summary

This PR adds a theme-level mitigation for the circular `article` shortcode render deadlock described in #2791.

When two articles embed each other with the `article` shortcode and summaries are enabled, Hugo can hang during rendering. This change detects the circular render path and safely degrades only the inner cyclic reference, while preserving normal behavior for non-cyclic references.

## Reproduction

Two articles that reference each other:

```md
<!-- content/posts/a/index.md -->
{{< article link="/posts/b/" showSummary=true >}}

<!-- content/posts/b/index.md -->
{{< article link="/posts/a/" showSummary=true >}}
```

Without this change, rendering `$target.Summary` causes Hugo to render `$target.Content`, which executes nested shortcodes. In a mutual reference, this can attempt to render content that is already being rendered and cause the build to hang.

### Add circular render detection

`layouts/shortcodes/article.html` now uses page-local `.Scratch` state to mark the page currently being rendered.

Before rendering an article card, the shortcode checks whether the target page is already marked as rendering. When a circular reference is detected, it passes a `cyclic` flag to the article-link partial and disables the summary for that inner reference.

### Skip content-dependent output on cyclic references

`layouts/partials/article-link/_shortcode.html` skips both the summary and metadata when `cyclic` is true.

Skipping metadata is necessary because `ReadingTime` and `WordCount` may also require Hugo to access the target page content, which can otherwise re-enter the same deadlocked render path.

The title, link, and feature image remain available on the cyclic article card.

### Respect explicit `showSummary=false`

The previous `default` chain treated `false` as an empty value. Therefore, an explicit shortcode option such as:

```md
{{< article link="/posts/b/" showSummary=false >}}
```

could be overridden by the page or site default.

This PR distinguishes between these cases:

- `showSummary` omitted: use the target page or site default;
- `showSummary=true`: show the summary;
- `showSummary=false`: explicitly disable the summary.

## Result

| Reference type | Summary | Metadata | Title / link / feature image |
| --- | --- | --- | --- |
| Non-cyclic reference | Preserved | Preserved | Preserved |
| Cyclic reference | Safely skipped for the inner reference | Safely skipped for the inner reference | Preserved |

## Verification

Tested with two articles that mutually reference each other through the `article` shortcode with `showSummary=true`.

- Before this change, the Hugo build timed out while rendering.
- After this change, the build completes successfully.
- Both generated pages contain the expected `article-link--shortcode` card.
- A shortcode with `showSummary` omitted still inherits the configured site default.

### DIFF
--- a/layouts/partials/article-link/_shortcode.html
+++ b/layouts/partials/article-link/_shortcode.html
@@ -9,7 +9,9 @@
 */}}
 {{ $target := .target }}
 {{ $shortcodeShowSummary := .showSummary }}
+{{ $shortcodeShowSummarySet := .showSummarySet }}
 {{ $shortcodeCompactSummary := .compactSummary }}
+{{ $cyclic := .cyclic | default false }}
 {{ $constrainItemsWidth := site.Params.list.constrainItemsWidth | default false }}
 {{ $disableImageOptimization := site.Store.Get "disableImageOptimization" }}
 
@@ -120,10 +122,19 @@
         {{ partial "extend-article-link.html" $target }}
       {{ end }}
     </header>
+    {{ if not $cyclic }}
     <div class="text-sm text-neutral-500 dark:text-neutral-400">
       {{ partial "article-meta/basic.html" $target }}
     </div>
-    {{ if $shortcodeShowSummary | default $target.Params.showSummary | default site.Params.list.showSummary | default false }}
+    {{ end }}
+    {{/* Respect an explicit showSummary=false from the shortcode; only fall
+         back to target/site defaults when the shortcode did not set it. The
+         plain `default` chain treats false as zero and would override it. */}}
+    {{ $summaryEnabled := $shortcodeShowSummary }}
+    {{ if not $shortcodeShowSummarySet }}
+      {{ $summaryEnabled = $target.Params.showSummary | default site.Params.list.showSummary | default false }}
+    {{ end }}
+    {{ if $summaryEnabled }}
       {{ $compactSummary := $shortcodeCompactSummary | default $target.Params.compactSummary | default false }}
       <div
         class="article-link__summary prose dark:prose-invert max-w-fit mt-1 {{ if $compactSummary -}}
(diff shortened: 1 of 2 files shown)


ZhenShuo2021:
Since Hugo renders in parallel, this fix makes circular references hide summaries in a way that's non-deterministic, depends on render order, not something content authors can control from markdown. This makes builds not reproducible. The same content might produce different output locally vs in CI. Different machines, different levels of parallelism, and different Hugo versions may all produce different results.

friday-oc:
Thanks for the careful investigation and reproduction. We’re going to close this version because the Scratch-based cycle detection makes the rendered result depend on Hugo’s parallel render order, so the same content can produce different output across runs or environments. We need a deterministic solution before changing the article shortcode’s behaviour. Please feel free to reopen the discussion with an approach that avoids render-order-dependent output.

PR #3002, opened 2026-07-10
New Shortcode: Add stats for display of (number,label) tuples 

richtefee:
This PR adds a new stats shortcode with a nested stat sub-shortcode to render responsive statistic cards.

Features:
- Supports configurable grid columns via max-cols
- Supports cycling through custom theme colors via the colors parameter
- Allows each statistic to define a number and label

The color coding scheme seems suboptimal. Feedback on this would be appreciated.

### DIFF
--- a/exampleSite/content/docs/shortcodes/index.md
+++ b/exampleSite/content/docs/shortcodes/index.md
@@ -926,6 +926,64 @@ You can see some additional Mermaid examples on the [diagrams and flowcharts sam
 
 <br/><br/><br/>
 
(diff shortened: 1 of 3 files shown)


friday-oc:
Thanks for the contribution and for flagging the visual-design questions in the PR. We’re going to close this version because the shortcode needs a clearer design and integration approach before it fits the theme. In particular, the global CSS, inline colour construction, and validation behaviour need more work. A discussion first would be the best route for a future version.

PR #3004, opened 2026-07-11
feat(header): add mobile submenu checked config

yozlog:
This PR is a follow-up to #2980 and contains only the mobile menu changes requested during review.

Regarding the question about why it isn't collapsed by default, my reasoning was that, in some scenarios, the sub-items are not particularly important. Having them expanded by default can make the menu feel a bit cluttered.

## Summary

Set the default appearance of the mobile menu as shown in the image below.

mobile-menu

## Changes

- Add a new parameter to `params.toml` to control whether nested menu items on mobile devices are expanded by default.
- Update the mobile menu behavior to respect the new configuration.
- Document the new parameter in the configuration docs.

## Notes

I'm not sure whether i should update the configuration docs or not, but I did anyway. 

Only the English and Simplified Chinese versions of the configuration docs have been reviewed. The other translations were generated using AI, and I cannot guarantee their grammatical or semantic accuracy.

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -54,6 +54,7 @@ forgejoDefaultServer = "https://v11.next.forgejo.org"
 
 [header]
   layout = "basic" # valid options: basic, fixed, fixed-fill, fixed-gradient, fixed-fill-blur
+  # mobileSubmenuChecked = true # set to false to collapse nested mobile menu items by default
 
 [footer]
   showMenu = true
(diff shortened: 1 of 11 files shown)


friday-oc:
Thanks for the follow-up work on the mobile submenu. We’re closing this version because the proposed  handling prevents an explicit  configuration from working, and the generated documentation changes add duplicated or unreviewed content. A smaller, corrected implementation can be proposed later if we decide to take this configuration option forward.

friday-oc:
Clarifying the configuration issue: using Hugo `default true` means an explicit `mobileSubmenuChecked = false` is treated as empty and cannot disable the expansion.

PR #3006, opened 2026-07-11
fix(a11y): add mastodon share labels

yozlog:
## Summary

- Fixes an accessibility issue where sharing links lack a discernible name when translations are missing.
- Adds missing `sharing.mastodon` translations across all language files.

## Changes

- **`layouts/partials/sharing-links.html`**: Added a fallback generator formatted as "Share via [Service]" if the translation is missing.
- **`i18n/*.yaml`**: Added `sharing.mastodon` key-value pairs to 31 files.

## Validation

- Verified that all sharing links expose a discernible `aria-label` under a11y checks.

### DIFF
--- a/i18n/ar.yaml
+++ b/i18n/ar.yaml
@@ -73,6 +73,7 @@ sharing:
   pinterest: "النشر على بينترست"
   reddit: "النشر على ريديت"
   twitter: "النشر على تويتر"
+  mastodon: "Share via Mastodon"
 
 shortcode:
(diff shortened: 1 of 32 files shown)


friday-oc:
Thanks for addressing the missing accessible names, the fallback mechanism is a useful idea. We’re closing this version because most of the added locale values are English text in non-English translation files. We would prefer either a fallback-only fix or proper reviewed translations, so the change does not degrade localisation quality.

Issue #2959, opened 2026-05-26
[Bug]: Taxonomies other than 'categories' and 'tags' are not shown as badges


schaerfo:
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Hugo Version Check

- [x] My Hugo version is within the supported range

### Describe the Bug

Only the taxonomies 'categories' and 'tags' can be displayed as badges in the article header, the home page, lists and such. Other (custom) taxonomies such as 'topics' cannot be shown and there is no config option to show them.

### To Reproduce

1. Create a taxonomy `topic = "topics"` in hugo.toml
2. Set `showTaxonomies` to `true`
3. Set `topics = ["other"]` for an article

### Expected Behaviour

On the page of the article a badge with the label 'other' should be displayed.

### Platform

- OS: Linux
- Browser: Firefox
- Version: 151.0

### Hugo Version

```console
hugo v0.161.1+extended+withdeploy linux/amd64 BuildDate=unknown
```

### Blowfish Version

```console
v2.103.0
```

### Additional Context

The bug was introduced in PR #2852.

The documentation states

> Blowfish has full support for Hugo taxonomies and will adapt to any taxonomy set up.

To my understanding, this implies users should be able to display taxonomies as badges, regardless of what they are named. This was also possible in versions 2.101 and prior. So we have an incompatible change. While it would be possible for users to rename their taxonomies to conform to Blowfish's expectations, this would invalidate existing links, something that users (including myself) would probably want to avoid.

A possible fix would be

```
diff --git a/layouts/partials/article-meta/basic.html b/layouts/partials/article-meta/basic.html
index d83c703d..28333f96 100644
--- a/layouts/partials/article-meta/basic.html
+++ b/layouts/partials/article-meta/basic.html
@@ -109,7 +109,7 @@
               {{ end }}
             {{ end }}
             {{ if .Params.showTags | default (.Site.Params.article.showTags | default true) }}
-              {{ if (eq $taxonomy "tags") }}
+              {{ if not (eq $taxonomy "categories") }}
                 <a class="relative mt-[0.5rem] me-2" href="{{ .RelPermalink }}">
                   {{ partial "badge.html" .LinkTitle }}
                 </a>
```
[...]

PR #3007, opened 2026-07-12
:bug: Fixed custom taxonomies not appearing as badges

MaikelChan:
After the PR https://github.com/nunocoracao/blowfish/pull/2852 I introduced a bug that caused that custom taxonomies would not appear as badges in an article. This was reported in issue https://github.com/nunocoracao/blowfish/issues/2959.

The proposed solution in the issue had another problem: `authors` and `series` taxonomies would appear duplicated because those are already handled in other places in the code.

I've implemented a fix that should fix all cases. The screenshot shows an article that has category, tags, authors, series and a custom taxonomy:

image

As seen, the authors taxonomies are not duplicated, series don't appear as badges, as it already appears in the series list below, and custom taxonomies are shown.

Also, custom taxonomies will show as long as the `showTaxonomies` parameter is true. There's some extra granularity for categories and tags with the `showCategories` and `showTags` parameters, but maybe it would be too much to have more parameters for any kind of taxonomy. For now, I just decided to leave it like this so the bug is fixed. At some other point it can be considered if it's necessary to add extra granularity for custom taxonomies.

### DIFF
--- a/layouts/partials/article-meta/basic.html
+++ b/layouts/partials/article-meta/basic.html
@@ -100,20 +100,26 @@
       {{ range $taxonomy, $terms := .Site.Taxonomies }}
         {{ if (gt (len ($context.GetTerms $taxonomy)) 0) }}
           {{ range $context.GetTerms $taxonomy }}
-            {{ if .Params.showCategories | default (.Site.Params.article.showCategories | default true) }}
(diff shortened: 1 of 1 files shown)


schaerfo:
I can confirm that this PR does indeed fix the issue I raised

PR #3014, opened 2026-07-18
Add reading progress bar for articles

mortennordbye:
## Describe your changes

Adds an optional reading progress bar for articles: a thin, primary-colored bar fixed to the top of the viewport that fills from 0 to 100 percent as the reader scrolls through a page.

**Why**

Long-form articles have no scroll-position indicator today. This is a common, lightweight blog-theme affordance and there is no equivalent in the theme (the scroll-to-top button only appears past a threshold and shows no progress).

**What**

- New `article.showReadingProgress` param (default `false`), overridable per page via front matter, following the same pattern as the other `article.show*` flags.
- New `layouts/partials/reading-progress.html` partial and `assets/js/reading-progress.js`, mirroring the existing scroll-to-top feature.
- The script is loaded as its own fingerprinted `<script>` tag (the same approach used for `zen-mode.js`) rather than appended to the shared `main.bundle.js`, because that bundle is cached by its target name and the inclusion is page-conditional.
- The bar sits at `z-100` so it renders above the fixed header while still being covered by the search overlay (`z-500`).
- Enabled on the exampleSite to showcase it, and documented in the configuration and front matter tables.

**Verification**

- `hugo` production build of the exampleSite succeeds with no errors; the script tag and bar element are emitted on article pages and absent on the home and list pages.
- Verified in the browser that the bar tracks scroll exactly (0 percent at top, 50 percent at mid, 100 percent at bottom) and paints above the fixed header, in both light and dark mode.
- Only the `.h-1` utility was added to the committed compiled CSS (via `npm run build`); no other CSS changed.

N/A

- [x] Will this be part of a product update? If yes, please write one phrase about this update.

Optional reading progress bar for articles, enabled with `article.showReadingProgress`.

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -1079,6 +1079,9 @@
   .h-0 {
     height: 0;
   }
+  .h-1 {
+    height: var(--spacing);
+  }
   .h-1\/2 {
     height: calc(1 / 2 * 100%);
   }
--- a/assets/js/reading-progress.js
+++ b/assets/js/reading-progress.js
@@ -1,0 +1,8 @@
+function readingProgress() {
+  const bar = document.getElementById("reading-progress");
+  const scrollable = document.documentElement.scrollHeight - window.innerHeight;
+  bar.style.width = scrollable > 0 ? `${(window.scrollY / scrollable) * 100}%` : "0%";
+}
+
+window.addEventListener("scroll", readingProgress);
+window.addEventListener("load", readingProgress);
--- a/assets/js/reading-progress.js
+++ b/assets/js/reading-progress.js
@@ -1,8 +1,9 @@
 function readingProgress() {
   const bar = document.getElementById("reading-progress");
+  if (!bar) return;
   const scrollable = document.documentElement.scrollHeight - window.innerHeight;
   bar.style.width = scrollable > 0 ? `${(window.scrollY / scrollable) * 100}%` : "0%";
 }
 
-window.addEventListener("scroll", readingProgress);
+window.addEventListener("scroll", readingProgress, { passive: true });
 window.addEventListener("load", readingProgress);
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -96,6 +96,7 @@ forgejoDefaultServer = "https://v11.next.forgejo.org"
   showPagination = true
   invertPagination = false
   showReadingTime = true
+  showReadingProgress = false
   showTableOfContents = false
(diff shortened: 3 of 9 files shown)


friday-oc:
Thanks, this is a useful and well-scoped optional feature. Before we merge, could you adjust the configuration handling so a page-level false value genuinely overrides a site-level true value? Hugo `default` treats false as empty, so the current condition cannot disable the bar per page once enabled globally. Please use an explicit parameter-presence check, for example initialise from the site setting, then use `isset .Params "showReadingProgress"` to assign the page value when it is present. Please apply the same resolved boolean in both `single.html` and `head.html`. Once that is fixed, this should be ready to merge.

mortennordbye:
@friday-oc thanks for the review, I've tried to match what you asked for.

The value is now initialised from the site setting and only overridden when the front matter parameter is actually present, so a page-level `false` genuinely disables the bar:

```go-html-template
{{ $showReadingProgress := site.Params.article.showReadingProgress | default false }}
{{ if isset .Params "showreadingprogress" }}
  {{ $showReadingProgress = .Params.showReadingProgress }}
{{ end }}
```

The same resolved boolean is used in both `single.html` and `head.html`, so the bar and its script can't get out of sync. I used the lowercase key in `isset` to match the existing usage in `language-redirect.html` and `analytics/fathom.html`.

Tested against the exampleSite with all four combinations (front matter `true` / `false` / absent, with the site setting both on and off), checking for the bar element and the fingerprinted script on each page:

| front matter | site `true` | site `false` |
| --- | --- | --- |
| `showReadingProgress: true` | shown | shown |
| `showReadingProgress: false` | hidden | hidden |
| absent | shown | hidden |

List pages emit neither, and the full multilingual build is clean. Let me know if you'd like anything else changed.

mortennordbye:
Pushed two small robustness fixes to the script while I was in here:

- Bail out when the progress bar element is missing. The script loads in the head without `defer`, so the scroll listener is registered before the element exists in the body. A scroll event fired during parsing, for example when the browser restores the scroll position on reload, would have thrown. This matches the guard already in `background-blur.js`. `scroll-to-top.js` has the same shape but ships in the deferred bundle, so it is not exposed to this.
- Marked the scroll listener `{ passive: true }`, as `toc.html` already does.

I left the update itself unthrottled to stay consistent with `scroll-to-top.js` and `background-blur.js`. Happy to switch it to `transform: scaleX()` inside a `requestAnimationFrame` if you would prefer that here.

PR #3024, opened 2026-07-26
feat: allow single pages without TOC to use full content width

drupol:
This PR updates the single article layout so pages without a visible table of contents can use the full available content width.

Currently, `single.html` always applies `max-w-prose` to the article header, content, and footer, and the main content wrapper always uses `max-w-fit`. That works well when a table of contents is displayed, but unnecessarily constrains pages where the TOC is disabled or not generated.

The change computes the TOC visibility once near the top of the template and reuses it to set the layout width:

- when the TOC is shown, the existing constrained `max-w-prose` / `max-w-fit` behavior is preserved;
- when the TOC is not shown, the header, article content, footer, and content wrapper can expand to `max-w-full w-full`.

This makes single pages without a TOC better suited for wider content such as custom shortcodes, media, embeds, tables, or layouts that should not be limited to prose width.

No issue.

 - [ ] Will this be part of a product update? If yes, please write one phrase about this update.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -1,5 +1,9 @@
 {{ define "main" }}
   {{ .Scratch.Set "scope" "single" }}
+  {{ $enableToc := site.Params.article.showTableOfContents | default false }}
+  {{ $enableToc = .Params.showTableOfContents | default $enableToc }}
+  {{ $showToc := and $enableToc (in .TableOfContents "<ul") }}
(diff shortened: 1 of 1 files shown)


friday-oc:
Thanks for the thoughtful implementation. We’re going to close this PR because it changes the default width of every single page without a table of contents, which is a broader product and design decision than we want to make without prior discussion. Please feel free to open a discussion first if you would like to revisit the layout direction.

Issue #3033, opened 2026-08-01
[Bug]: Site author is included when using multi-author support


Taffer:
### Issue Check

- [x] I have checked existing Issues and I feel this bug has not been raised

### Hugo Version Check

- [x] My Hugo version is within the supported range

### Describe the Bug

When using a page with [Multiple Authors](https://blowfish.page/docs/multi-author/), the `author` data set in the site params is *also* included.

### To Reproduce

In site params, set up an author:

```yaml
params:
  author:
    name: 'Taffer'
    fediverse_creator: '@Taffer@mastodon.gamedev.place'
    links:
      - mastodon: 'https://mastodon.gamedev.place/@Taffer'
```

Set up `data/authors/*.json` files per the [Multiple Authors](https://blowfish.page/docs/multi-author/) doc, say `author1.json` and `taffer.json`.

In an article, configure it to show that `author1` is the author:

```md
---
title: 'Some Article'
date: '2026-07-31T06:37:26-04:00'
authors:
    - 'author1'
showAuthor: true
showAuthorsBadges: true
---
```

As expected, the page shows the author chunk for `author1`, but `taffer`'s chunk is also there due to being set in the site params.

As a work-around, you can remove the `author:` section from the site params, but then you lose all the bits that depend on `.Site.Params.author` throughout the theme.

### Expected Behaviour

Only authors listed in the `authors:` are included in the rendered HTML.

### Platform

- OS: Linux
- Browser: Vivaldi
- Version: 8.1.4087.61 (Official Build) (64-bit)

### Hugo Version

```console
hugo v0.164.0+extended+withdeploy linux/amd64 BuildDate=unknown
```

### Blowfish Version

```console
v2.104.0
```

### Additional Context

This might be intentional behaviour?

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Taffer:
And a PR for you: https://github.com/nunocoracao/blowfish/pull/3034

nunocoracao:
Fixed by #3034 — the site default author is no longer rendered when a page sets `authors`, matching the documented override behavior. Ships in the next release. Thanks @Taffer!

PR #3034, opened 2026-08-01
Address #3033 - suppress default author when using multi-author

Taffer:
If a page has `authors` set, use that instead of the site's `author` parameter, otherwise use the `author` parameter.

Without the whitespace:

```sh
blowfish on  multi-author-no-default [!] is 📦 v2.105.0 via 🐹 v1.26.5-X:nodwarf5 via  v26.5.0
🐟  chrish@moon ❯ git diff --ignore-all-space main
diff --git a/layouts/_default/single.html b/layouts/_default/single.html
index 95389091..843e0b0c 100644
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -109,11 +109,7 @@
     {{ $baseURL = delimit (slice $baseURL "/") "" }}
   {{ end }}

-  {{ if .Params.showAuthor | default (site.Params.article.showAuthor | default true) }}
-    {{ $showAuthor = 1 }}
-    {{ partial "author.html" . }}
-  {{ end }}
-
+  {{ if .Page.Params.authors }}
     {{ range $author := .Page.Params.authors }}
       {{ $authorData := index $authorsData $author }}
       {{- if $authorData -}}
@@ -133,6 +129,12 @@
         {{ partial "author-extra.html" (dict "context" . "data" $authorData "link" $finalLink) }}
       {{- end -}}
     {{ end }}
+  {{ else }}
+    {{ if .Params.showAuthor | default (site.Params.article.showAuthor | default true) }}
+      {{ $showAuthor = 1 }}
+      {{ partial "author.html" . }}
+    {{ end }}
+  {{ end }}

   {{ if or $taxonomyLink $showAuthor }}
     <div class="{{ cond $isAuthorBottom "mb-10" "mb-5" }}"></div>
```

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -109,29 +109,31 @@
     {{ $baseURL = delimit (slice $baseURL "/") "" }}
   {{ end }}
 
-  {{ if .Params.showAuthor | default (site.Params.article.showAuthor | default true) }}
-    {{ $showAuthor = 1 }}
-    {{ partial "author.html" . }}
-  {{ end }}
-
-  {{ range $author := .Page.Params.authors }}
-    {{ $authorData := index $authorsData $author }}
-    {{- if $authorData -}}
-      {{ range $taxonomyname, $taxonomy := $taxonomies }}
-        {{ if (eq $taxonomyname $author) }}
-          {{ $taxonomyLink = delimit (slice $baseURL "authors/" $author "/") "" }}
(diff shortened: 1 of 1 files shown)


nunocoracao:
Thanks @Taffer for the clean fix (and for filing #3033 with a solid repro) — verified the demo's multi-author page now renders only the listed authors, matching the documented `authors` override behavior. Ships in the next release. 🙏

PR #3039, opened 2026-08-08
Add github_gist icon and social link support

0r1k:


nunocoracao:
Thanks @0r1k! Merged via #3047 with your commit credited (Co-authored-by) — trimmed to the icon and the commented config example, since we prefer not to add the gist link to the demo site's author links. The `github_gist` key works as of the next release.

PR #3041, opened 2026-08-09
Improve Largest Contentful Paint (LCP) performance for homepage profile images.

EricGusmao:
- Set `loading="eager"` to prevent lazy loading.
  - Set `fetchpriority="high"` to prioritize the LCP resource.
  - Apply the optimization to the `profile`, `hero`, and `background` homepage layouts.

### DIFF
--- a/layouts/partials/home/background.html
+++ b/layouts/partials/home/background.html
@@ -62,6 +62,8 @@
                 width="144"
                 height="144"
                 alt="{{ $.Site.Params.Author.name | default `Author` }}"
+                loading="eager"
+                fetchpriority="high"
                 src="{{ $final.RelPermalink }}"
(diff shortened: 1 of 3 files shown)


nunocoracao:
Thanks @EricGusmao, merged! 🙏 This is spot-on for the `profile` layout, where the avatar is the LCP element.

If you're up for a follow-up PR, there's a natural next step: on the `background` and `hero` layouts the LCP element is actually the full-viewport homepage image (the first `` in `layouts/partials/home/background.html` and `hero.html`), so `fetchpriority="high"` belongs on that one there — and `card.html`'s homepage image could use the same treatment. Ideally the avatar would keep default priority on those layouts so it doesn't compete with the LCP resource. (Minor: `loading="eager"` is the browser default, so those lines are optional.)

Happy to review if you send it!

PR #3066, opened 2026-08-21
Fix KaTeX not rendering due to shared bundle cache collision

0r1k:
katex-render.js was the only per-page conditional input to the resources.Concat "js/main.bundle.js" target. Hugo caches Concat output by target path, so whichever page Hugo processes first for that path determines the bundle for every page on the site - silently dropping the katex render call on pages that need it whenever a non-katex page is built first.


0r1k:
Closing in favor of #3063, which fixes the same root cause the same way - no need for two PRs.

PR #3072, opened 2026-08-26
🐛 Render HTML in series widget titles

tekenstam:
## Summary
The series widget printed `{{ $post.Params.title }}`, which Hugo HTML-escapes. Front-matter titles that include markup (for example `<br>` and `<small>` for a subtitle) showed as literal tags in the series list, while the article heading and prev/next links already used `{{ .Title | emojify }}` and rendered correctly.

## Change
Use `{{ $post.Title | emojify }}` in `layouts/partials/series/series_base.html` so series links match the rest of the theme.

## Test plan
- [x] Add HTML to a series post `title` in the example site (or any site using series)
- [x] Confirm the series widget renders the markup instead of showing raw tags
- [x] Confirm plain-text titles are unchanged

### DIFF
--- a/layouts/partials/series/series_base.html
+++ b/layouts/partials/series/series_base.html
@@ -17,7 +17,7 @@
         class="py-1 border-dotted border-neutral-300 border-s-1 -ms-5 ps-5 dark:border-neutral-600">
         <a href="{{ $post.RelPermalink }}">
           {{ i18n "article.part" }} {{ $post.Params.series_order }}:
-          {{ $post.Params.title }}
+          {{ $post.Title | emojify }}
         </a>
       </div>
     {{ end }}


tekenstam:
If you open the series widget of this post you will see how the fix behaves (I applied it as a local override).
https://todd.ekenstam.dev/posts/202608-simplifying-kubernetes-part-1/

Or you can reproduce it by creating a series with a title like this:
```
title: "The Developer Experience Problem with Kubernetes at Scale<br><small>Simplifying Kubernetes, Part 1</small>"
series:
  - Simplifying Kubernetes
series_order: 1
```

drupol:
Just stumbling across this issue and I am now wondering why this series list of links is not in a proper `<ul><li>` structure.

drupol:
@nunocoracao Would it be a good idea to convert the series links using `<ul>,<li>` ?

